PR #869, opened 2015-01-03
Add PetSpectAnalysis extension. This extension will contain modules to p...

gti-fing:
...erform analysis based on Pet and Spect images. Currently, there is one module which allows for the quantification of glucose uptake using dynamic PET brain scans.  Voxel or region based Patlak analysis can be performed . Input function estimation can be performed using image derived and population derived  methods.


gti-fing:
Iconurl and screenshoturls were set to valid images.
Thanks!

PR #889, opened 2015-01-19
ENH: SlicerRT 0.15: Increase version number to indicate completeness of ...

cpinter:
...DICOM export; Fixes in structure set loading. https://www.assembla.com/code/slicerrt/subversion/changesets/compare/2068...2095


cpinter:
Thanks!

PR #901, opened 2015-02-06
ENH: SlicerRT 0.15.1: Fixed build issue due to Slicer core change; Use u...

cpinter:
...pdated gamma comparison; Support DICOM references for RT objects. See more info at https://www.assembla.com/code/slicerrt/subversion/changesets/compare/2095...2136


cpinter:
Thanks!

PR #902, opened 2015-02-06
ENH: GelDosimetryAnalysis 0.2.2: Use updated gamma comparison; Added lin...

cpinter:
...e profile tool; Export calibration curve to CSV; Updated registration parameters to support new datasets. See more info at https://www.assembla.com/code/slicerrt/subversion/changesets/compare/2049...2138


cpinter:
Thanks, Andrey!

PR #903, opened 2015-02-07
BUG: Fixed MultidimData build error

lassoan:
These errors recently showed up on the dsahboard: error: no viable conversion from 'QByteArray' to 'const std::string' (aka 'const basic_string')
This pull request should fix this problem.

### DIFF
--- a/MultidimData.s4ext
+++ b/MultidimData.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm svn
 scmurl https://subversion.assembla.com/svn/slicerrt/trunk/MultidimData/src
-scmrevision 1980
+scmrevision 2140
 
 # list dependencies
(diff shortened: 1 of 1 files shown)


jcfr:
Talking about similar issue with collegue .. I learn that using `qPrintable()` should also work.

lassoan:
qPrintable is a nice shortcut and we should use it for sending it to stdout or stderr.

We should add a similar macro or function for converting QString to const char* that is intended to be used in VTK, because it allows us to do all encodings consistently (toLatin1/toLocal8Bit/other) and when character encoding will be improved in VTK then we could review/change these conversions more easily. There is already vtkUnicodeString that is used in text rendering and file IO and I guess unicode strings may gradually be used in more and more places.

jcfr:
Fully agree.

PR #906, opened 2015-02-25
ENH: SlicerRT 0.15.2: Automatic oversampling calculation for contour ras...

cpinter:
...terization (and so DVH) added; Improved ribbon model generation; Fixed tests. See more info at https://www.assembla.com/code/slicerrt/subversion/changesets/compare/2136...2171


cpinter:
Thanks!

PR #917, opened 2015-03-11
ENH: Add DeveloperToolsForExtensions

fbudin69500:
This extension exposes some functions that are useful for developers. So far, it allows to install an extension directly from its archive (*.tar.gz or *.zip). This is useful to verify that an extension packages correctly. It could also be used to disseminate extensions directly from a laboratory's website or a personal webpage. It also allows a user to load a module into Slicer while Slicer is already running.


jcfr:
Thanks for your contribution :+1:

PR #927, opened 2015-03-28
Add Sequences module

ihnorton:
(formerly Multidim Data in SlicerRT) per @lassoan this is moved from SlicerRT to a separate module on GitHub for future development.


lassoan:
I'll merge this now and if it works well then Sequences will be moved to the Slicer core as a remote module.

PR #937, opened 2015-04-03
BUG: fix logging

naucoin:
Needed to import the logging package for the round trip test to pass.
Fixes: ReportingRoundTripTest

### DIFF
--- a/Reporting.s4ext
+++ b/Reporting.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl git://github.com/fedorov/Reporting.git
-scmrevision 5fc8b8c
+scmrevision 3230524
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files
--- a/Reporting.s4ext
+++ b/Reporting.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl git://github.com/fedorov/Reporting.git
-scmrevision 3230524
+scmrevision e787f9b
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files


naucoin:
Fixed the DICOM Seg tests.
The Reporting Round Trip test is failing due to a crash on exit tracked back to the Data Probe module parameter node, it needs to be debugged outside of Reporting, so the current commits should be good to merge for now.

fedorov:
@naucoin just wanted to point out that the convention for the topic names for ExtensionIndex updates is "update <ExtensionName>". Sorry I was too busy when you submitted and didn't have a chance to tel you about this or update the topic.

jcfr:
See http://wiki.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/Tutorials/Contribute_Extension_Description_File#Requesting_an_extension_update

jcfr:
I guess, we shouldn't prevent user from having more explicit topic name if they want.

fedorov:
I agree, but then it should be something  like
```
update Reporting: fix logging
```
anyway, just wanted to help. I am fine either way.

jcfr:
Are we talking about the topic name or the commit message ?

For the topic name, would prefer not having spaces or semicolons.

Instead of 
```
update Reporting: fix logging
```

 the following would be better (and machine friendly)

```
update-Reporting_fix-logging
```

fedorov:
I was talking about the title of the pull request that shows on top of this page

jcfr:
I see.  Makes sense in that case.

naucoin:
Ah, sorry about the confusion, I was trying to be clear and not re-use a former branch name, I didn't notice  that the pull request would take the commit message rather than the branch name.

PR #942, opened 2015-04-13
Add T1_Mapping

stevedaxiao:
I created a T1_Mapping extension description file.
Please let me know if you have any question.

Thanks

### DIFF
--- a/T1_Mapping.s4ext
+++ b/T1_Mapping.s4ext
@@ -1,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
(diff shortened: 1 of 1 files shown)


jcfr:
The extension fail to build. See http://slicer.cdash.org/index.php?project=Slicer4&filtercount=1&showfilters=1&field1=buildname/string&compare1=63&value1=T1_Mapping&date=2015-04-15

This is because there are no `CMakeLists.txt` file in the top-level folder. 

Two options:

1. fix the extension build system to look for a `CMakeLists.txt` file in child directories.
2. update your extension adding a `CMakeLists.txt` at the top. See below for an example.
3. update your extension by re-organization the files. Basically more all the files one folder up.

```
cmake_minimum_required(VERSION 2.8.9)
project(T1_Mapping)
add_subdirectory(T1_Mapping)
```

I would suggest option (3)

stevedaxiao:
Got it. Will do that. Thank you.

Xiao

PR #948, opened 2015-04-29
ENH: Updated Extension for UKF.

hjmjohnson:
@jcfr I don't have permission to push directly.  Here is the patch to test for UKF.

Hans


jcfr:
Please consider also updating the file associated with https://github.com/Slicer/ExtensionsIndex/tree/4.4

PR #955, opened 2015-05-12
Update TrackerStabilizer

lchauvin:
Changed code structure, slightly improved UI, added a MRML node and modified filtering to avoid gimbal lock (Andras)

https://github.com/SlicerIGT/TrackerStabilizer/compare/cae960ae5c2...master

### DIFF
--- a/TrackerStabilizer.s4ext
+++ b/TrackerStabilizer.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl git://github.com/SlicerIGT/TrackerStabilizer.git
-scmrevision cae960ae5c21b4a238781bf17a1421a0e56ca1a4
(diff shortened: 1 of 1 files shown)


jcfr:
FYI - There is a Slerp function in vtkQuaternion.
See https://github.com/Kitware/VTK/blob/cdbec3e4f1dfe74556ddf27580b44a86d2ebcc94/Common/Math/vtkQuaternion.h#L190

Would that be helpful ?

jcfr:
Note for later: Make sure to create a branch named `update-TrackerStabilizer` instead of creating the PR directly of master.

lchauvin:
Oh yes, sorry I forgot. Thanks for reminding me.

lchauvin:
Oh yes you're right. I will look into it, and if I got same results (which should be the case), maybe replace it.

Thank you.

PR #960, opened 2015-05-16
update T1_Mapping.s4ext

stevedaxiao:
Dear Slicer Developer,
Not similar to the previous version, the latest nightly version of Slicer doesn't do "import string" by default, I have to add this line and update my python module to let the module work. 
Also the batch mode should be added as well.

Thanks and best regards,

Xiao


jcfr:
I would recommend to explicitly import all python modules used in your module.

PR #1026, opened 2015-07-03
ENH: Add Q3DC extension

luciemac:
Description:
This extension contains one module of the same name. Using placed
fiducials, it allows users to compute 2D angles: Yaw, Pitch and Roll;
and decompose the 3D distance into the three different components: R-L ,
A-P and S-I. It is possible to compute the middle point between two
fiducials and export the values.

Contributors:
Lucie Macron (University of Michigan)


fbudin69500:
@luciemac, can you change the icon url (which is still pointing out to the example fake one)?

PR #1047, opened 2015-07-16
Update IASEM extension

blowekamp:
346a715 Add additional test for streaming CLI
6feb00b Update IASEMImport and IASEMImportSeries to support streamed writing
75ba9c2 Update BinaryWatershed effect with split size parameters
da1ca32 Change BinaryWS effect to be based on Effect
0df235f Update BinaryWatershedFromMarker to new EditorLib module imports
882f3a1 COMP: Use vtkMRMLLabelMapVolumeNode

Thanks to Andras Lasso <lasso@queensu.ca> for a patch.

### DIFF
--- a/IASEM.s4ext
+++ b/IASEM.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl git://github.com/blowekamp/Slicer-IASEM.git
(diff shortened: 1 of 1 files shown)


blowekamp:
@pieper This patch contains fixes of the IASEM BinaryWatershed editor effect which I believe you were reporting to have errors.

PR #1059, opened 2015-08-10
ENH: SlicerRT nightly now builds the current master of its repository…

cpinter:
…, similarly to other extensions. This is to make releases less error-prone, and to make use of the automatic testing so that the developers are warned about regressions overnight, without manual interaction.

### DIFF
--- a/SlicerRT.s4ext
+++ b/SlicerRT.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm svn
 scmurl https://subversion.assembla.com/svn/slicerrt/trunk/SlicerRt/src
-scmrevision 2320
+scmrevision master
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files


jcfr:
:+1: Let's just make sure that before creating the branch `v4.5` of the extension index .. a fix SHA1 is added to the file.

cpinter:
I'm still planning to do the usual release process for the stable versions, see
https://www.assembla.com/spaces/slicerrt/wiki/SlicerRt_developers_page
What do you mean by SHA1 fix?

jcfr:
Great.

I just meant that after the next release of Slicer, the master of the index will become the new `v4.5` branch. 

During (Just before/or Just after) the Slicer release process, the` SlicerRT.s4ext` file would have to be updated to use a specific revision instead of `master`.

cpinter:
Got it, thanks! Will do for sure.

PR #1076, opened 2015-09-04
Add GraphCutSegment extension

DaphneCD:
This is a segmentation extension using Graph Cut and Star Shape algorithm.

I've added the copyright in the front of the header files under TumorSegm folder and cleared that the code is only for research use only.


jcfr:
Closing. Superseded by #1093

PR #1092, opened 2015-09-11
add T1Mapping

stevedaxiao:
Add T1Mapping newer version
scmurl corrected


stevedaxiao:
Thanks and T1Mapping should be ready.

PR #1096, opened 2015-09-16
add new extension MarginCalculator.

kevinwangcanada:


fedorov:
Is there a good reason why you are adding this extension to 4.4 branch, but not to the trunk?

cpinter:
Kevin told me the reason is that he has not made the changes necessary to make it compatible with the nightly (i.e. LabelMapVolumeNode), but he would like to publish it as soon as possible.

fedorov:
Sounds like a good reason for testing, even though somewhat confusing since new features are added to the release that are not available in the nightly :) I hope @jcfr will be ok with this...

cpinter:
I don't see any harm in adding it to a soon obsolete stable Slicer version, although I also think that releasing it for the nightly would be better. Thanks for merging!

fedorov:
Agreed.

btw @cpinter it looks like you should have the permissions to merge on ExtensionIndex as well - it looks like you are on the appropriate team.

PR #1111, opened 2015-10-05
BUG: This project used to be built as a superbuild project and theref…

fbudin69500:
…ore the build_subdirectory was set to the inner-build directory of the project. Because of problems using superbuild (reconfiguration, erasing CMake variables in the inner-build directory), this project does not explicitly use the superbuild system when compiling as a Slicer extension. the variable "build_directory" had not been updated in the description file.


jcfr:
:+1:

PR #1112, opened 2015-10-05
BUG: Remove new line characters in CMake variables

fbudin69500:
New line characters in CMake variables are creating problems. It was not possible to upload the package of the 3D Slicer extension on the Midas serv
er.

https://github.com/fbudin69500/MeshToLabelMap/compare/101bbadb07d57d57cb36ceea5e4f2e20bebdca37%E2%80%A6b0be1f0ded2d64535440755e1c4d0f7098e2cd81

### DIFF
--- a/MeshToLabelMap.s4ext
+++ b/MeshToLabelMap.s4ext
@@ -6,36 +6,36 @@
 
 # This is source code manager (i.e. svn)
 scm git
-scmurl https://github.com/NIRALUser/MeshToLabelMap.git
-scmrevision 101bbadb07d57d57cb36ceea5e4f2e20bebdca37
+scmurl git://github.com/NIRALUser/MeshToLabelMap.git
+scmrevision b0be1f0
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files
 # - The dependencies will be built first
 depends     NA
 
-# Inner build directory (default is .)
+# Inner build directory (default is ".")
 build_subdirectory .
 
 # homepage
-homepage   http://www.slicer.org/slicerWiki/index.php/Documentation/Nightly/Extensions/MeshToLabelMap
+homepage    http://www.slicer.org/slicerWiki/index.php/Documentation/Nightly/Extensions/MeshToLabelMap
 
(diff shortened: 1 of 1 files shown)


jcfr:
> New line characters in CMake variables are creating problems

Is it a problem during the upload ?

May be the function `midas_escape_url` needs to be fixed ?
See https://github.com/Slicer/Slicer/blob/3e3305fc5821f9b2d8a24d78c9c7e5cebf906580/Extensions/CMake/MIDASAPIUploadExtension.cmake#L98
and  https://github.com/midasplatform/Midas/blob/2f9ac1147d83e87f06c66fbadb4c664990c8f605/modules/packages/public/clients/cmake/MidasAPIScript.cmake#L58-L62

fbudin69500:
It was creating a wrong URL for the upload (with newlines in the URL).
I guess the function could indeed be improved.

fbudin69500:
It seems that this function in Slicer already handles quite a few special characters:
https://github.com/Slicer/Slicer/blob/3e3305fc5821f9b2d8a24d78c9c7e5cebf906580/CMake/MIDASAPILogin.cmake#L78-L108
Maybe it should also replace a new line by the LF character by %0A (I guess it is not necessary to keep CR [%0D] if it is a file from Windows?).
Or just detect it and through an error? How is the additional information (description, contributors,...) used on Midas? I don't see if directly on the website. Is it meta-information attached to the package?

jcfr:
> I guess it is not necessary to keep CR [%0D] if it is a file from Windows?

Indeed, keeping only LF will work

> Or just detect it and through an error? 

Having metadata with new line should be valid. This function should be fixed.

> How is the additional information (description, contributors,...) used on Midas? 

See https://github.com/midasplatform/slicerpackages/blob/c3c9a3e291df5a5454da44f4c50c2fea956e0d42/controllers/components/ApiComponent.php#L148-L296

> I don't see if directly on the website. Is it meta-information attached to the package?

It is both (1) attached to the package by adding the description file to the archive and (2) passed as metadata as parameter URL.

PR #1176, opened 2015-11-18
BUG:Update the name of some functions to support VTK6

pqpqpqpqpq:
Update the name of some functions to support VTK6
SetInput->SetInputData
SetInputPolyData->SetInputPolyDataConnection


jcfr:
Note that you would have to submit a request against the 4.5 branch to ensure your extension also compile and work with the latest stable

PR #1187, opened 2015-11-24
update QIICR extensions

che85:
* update version for mpReview (fix failing selftest import of the preprocessor module)


fedorov:
:+1:

fedorov:
Ah, I was a bit too fast with :+1: ;-)

For the next time, one small correction is that the title of the issue (and the branch which you are merging) should match the name of the extension you are updating, so in this case should be "update mpReview 4.5".

I had "update QIICR extensions" for the previous one because I updated several extensions at once.

I will edit the titles.

It is still :+1: but with a tiny caveat!

PR #1193, opened 2015-11-26
add Self Test for GraphCutSegmentExtension

DaphneCD:
Self Test is added into GraphCutSegmentExtension. The comparison of new commit and previous commit can be found here https://github.com/DaphneCD/GraphCutSegmentExtension/compare/479188c...2c4e572 .

I followed the instruction for Self Test Module to create the Self Test as a module. On my local build Slicer, I tried the Self Test module and it passed testing. But I'm not sure how to add the test into the extension. So I simply add the Python test file under Testing folder. The complete Self Test Module code can be found here https://github.com/DaphneCD/GraphCutSegmentSelfTest . Plase let me know if there is any problem.

Another thing is that there is an odd situation for my extension. It works totally fine on Mac. But on Windows it will have Slicer stop working problem with proper operation. The proper process is the work flow in Self Test. Hope it can help with finding the reason for Windows version crash problem.

Thanks!
Daphne

### DIFF
--- a/GraphCutSegment.s4ext
+++ b/GraphCutSegment.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl https://github.com/DaphneCD/GraphCutSegmentExtension.git
-scmrevision 479188c
+scmrevision 2c4e572
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files


pieper:
Hi Daphne - the self test module should be bundled included within your extension not a new repository - that is, the layout should be what you get if you use the extension wizard to create a new scripted module within your extension and then implemented the self test in that scripted module.  In that situation the self test will be executed as part of the build and test process on the factory on all platforms.

Having the self test that replicated the crash situation will definitely help with the debugging process.

HTH,
Steve

PR #1213, opened 2015-12-16
Fix many CMake problems

acil-bwh:


fedorov:
Please note that the convention for the issue title is "update <extension name>". You are updating ExtensionIndex repository, and "Fix many CMake problems" is not relevant for ExtensionIndex, this is the detail pertinent to your specific extension, which should be put in the description (empty in your commit).

PR #1289, opened 2016-03-11
ENH: Update AnglePlanesExtension extension

jbvimort:
This updates the AnglePlanesExtension extension to 5c0e76a04cb94e5b0c85ee9b82224e534b4bbd27.

See https://github.com/DCBIA-OrthoLab/AnglePlanes-Extension/compare/3e8fffa7b424c26a5ec2dac91614289ceb7cb221...5c0e76a04cb94e5b0c85ee9b82224e534b4bbd27 to view changes to the extension.

### DIFF
--- a/AnglePlanesExtension.s4ext
+++ b/AnglePlanesExtension.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl git://github.com/DCBIA-OrthoLab/AnglePlanes-Extension.git
-scmrevision 3e8fffa7b424c26a5ec2dac91614289ceb7cb221
+scmrevision 5c0e76a04cb94e5b0c85ee9b82224e534b4bbd27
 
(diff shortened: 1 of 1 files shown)


lassoan:
Hi Jean-Baptiste,

I see that you update your extensions a lot. To make the process simpler I would recommend two changes:
* Put all related modules in a single extension. It makes everyone's life easier. There are less clutter in the extension manager, on the dashboard, extension index; users can install all related modules with a single click, etc.
* Instead of using a specific git hash, use 'master'. Then the latest master version is used in the nightly build automatically and you don't need to send pull requests anymore for each change.

jbvimort:
Hello Andras,

I didn't knew that the second option was possible, but obviously it's a way better option than updated the extension index every week! Thank you for the information!
I just finished to change all my pull request to put the hash to 'master'!

Sorry for having bothering you so many times!

Have a good weekend!

PR #1295, opened 2016-03-11
ENH: Update CMFreg extension

jbvimort:
This updates the CMFreg extension to bbdc2f912cc86d4babc685bb1a177ebd61f5bf0c.

See https://github.com/DCBIA-OrthoLab/CMFreg/compare/c1d56dffa68a90dbb31905570a6898cb467c8594...bbdc2f912cc86d4babc685bb1a177ebd61f5bf0c to view changes to the extension.

### DIFF
--- a/CMFreg.s4ext
+++ b/CMFreg.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl https://github.com/DCBIA-OrthoLab/CMFreg.git
-scmrevision bbdc2f912cc86d4babc685bb1a177ebd61f5bf0c
+scmrevision master
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files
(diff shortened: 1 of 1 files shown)


fedorov:
You are updating your 4.5 release version to track master, the same as Slicer nightly. This is generally not a good idea, and I am not sure it is a good idea beyond in general. Slicer API changes, and you can/will get into incompatibility issues if you use the same code both for master and release. Can you confirm? I also want other folks to comment if it is acceptable to track master of the extension in the release branch of the ExtensionIndex. cc: @jcfr @pieper @lassoan

lassoan:
For 4.5 release version a separate branch should be created.

pieper:
Agreed

jbvimort:
So, just to be sure that I understand correctly: I need to create a branch (that I will probably name something like "4.5") in each repository and then put the name of this branch in the the s4ext of each extension?

fedorov:
This would be my recommendation.

Would also be helpful if you closed the current 4.5 PRs.

jbvimort:
Ok, sounds good to me!

Thank you so much for your advices!

jbvimort:
I finally change all the s4ext files to have two different branches for the nightly and the 4.5 stable release!

Thank you so much for all the informations!

PR #1309, opened 2016-04-20
update mpReview

fedorov:
fix build issue

### DIFF
--- a/mpReview.s4ext
+++ b/mpReview.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl https://github.com/SlicerProstate/mpReview.git
-scmrevision 51f40cc3917c02328d6d9dfc2ee5007813c9d687
+scmrevision 97b418bd6853c6f34bbd3a2fbf2c69f96225f80f
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files


fedorov:
@che85 FYI

che85:
@fedorov I guess we have some problems here since I called the python file of SlicerProstateUtils the same like the subdirectory living next to it. Modules are loaded randomly so it can happen, that mpReview is loaded after SlicerProstateUtils and then tries to import from the SlicerProstateUtils.py instead of looking in the subdirectory.

Any idea how we can deal with that? Maybe just renaming the python file?

fedorov:
Sorry, I don't exactly understand what is going on...

che85:
@fedorov When you take a look at https://github.com/SlicerProstate/SlicerProstate/tree/master/SlicerProstateUtils
you will notice, that the python file has the same naming like the subdirectory, which causes problems when the module has been loaded successfully by Slicer. If you would try to import then from that module it would try to import from the python file instead of looking in the subdirectory.

Example: from SlicerProstateUtils.mixins import ModuleWidgetMixin

This line would try to find mixins in the SlicerProstateUtils.py after SlicerProstateUtils has been loaded successfully in Slicer.

I hope that explains it a bit better?

fedorov:
I am at home today staying with my son, he hurt his leg... Cannot look at
the moment. Do you have a working solution? Sure, let's rename if that
solves the issue.

che85:
I will think about a good solution. I also don't like the fact, that we cannot just set the dependency to a project "SlicerProstate", that would be much easier.

fedorov:
But I thought now that you have a module, you should be able to set a
dependency!? I should be able to take a look around 2.

PR #1323, opened 2016-06-01
New Extension: ShapeQuantifier

jbvimort:
This extension is a collection of different modules that will enable the user to quantify shape differences between two models and/or compute shape statistics in a model .
ShapeQuantifier includes preprocessing tools, allowing the user to edit the models by cropping them (EasyClip) and/or making registration between models(SurfaceRegistration).
The user will have the choice to compute distances between points (Q3DC) or angles between lines (Q3DC) or planes (AnglePlanes).
He will be able to create distance maps (ModelToModelDistance) to either display them as color maps (ShapePopulationViewer) or compute statistics ( MeshStatistics) on the entire model or just a region of interest (PickAndPaint).
All those modules will be available to use by themselves or all at once by using ShapeQuantifier that will guide the user through the different steps.

Documentation: https://www.slicer.org/slicerWiki/index.php/Documentation/Nightly/Extensions/ShapeQuantifier

### DIFF
--- a/ShapeQuantifier.s4ext
+++ b/ShapeQuantifier.s4ext
@@ -1,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl git://github.com/jbvimort/ShapeQuantifierExtension.git
+scmrevision nightly
+
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends NA
+
+# Inner build directory (default is ".")
+build_subdirectory .
+
+# homepage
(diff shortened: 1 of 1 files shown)


jcfr:
Cc: @Slicer/slicer-core

ShapeQuantifier.s4ext
```diff
@@ -0,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl git://github.com/jbvimort/ShapeQuantifierExtension.git
+scmrevision nightly
```
jcfr: No problem doing so.

It just mean that you will have to change it to `4.6` after the next stable release of Slicer is done.

jbvimort:
Yes the plan was to create one branch for the nightly and one for the 4.5, and then continue in the same way with the next versions. That will avoid to bother you each time we want to push a small modification. :)

PR #1346, opened 2016-08-01
Add PETLiverUptakeMeasurement extension

chribaue:
PETLiverUptakeMeasurement extension is ready to be added to the stable release of Slicer.


fedorov:
@chribaue thank you for your contribution!

Looking at the wiki page, are you absolutely sure it is a good idea to have a separate extension for this? Why not add it as a module to PET quantitation extension, or PET segmentation?

PR #1366, opened 2016-11-16
ENH: Update NeedleFinder

mehrtash:
Updating needle finder module.
* [Compare Link](https://github.com/gpernelle/NeedleFinder/compare/6cf994c5e5f52d3c9bd6d6c97b737e6ed1a4b73a...34c4b18d2e65a3f33a392d3474be97b19b958557)


fedorov:
Just in case FYI - you can use `master` as your revision, this way you will not need to update extension file, but also means you need to be more careful about committing into master.

PR #1368, opened 2016-11-25
rename and update Reprting

fedorov:
now QuantitativeReporting

@jcfr I have few questions:

 - it looks like the updated wiki always has https URLs - will this be a
   problem? I put http prefix, but not sure if this is the right way to handle it.
   I recall in the past there were some issues with https somewhere.

 - is it allowable to have ` and ' in the s4ext metadata items? Just want to
   make sure I don't break the dashboard

 - let me know if you see I missed anything in the process of renaming an
   existing extension

### DIFF
--- a/QuantitativeReporting.s4ext
+++ b/QuantitativeReporting.s4ext
@@ -1,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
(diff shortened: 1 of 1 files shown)


fedorov:
@jcfr I am merging, because I am afraid I may forget, and I want to see if it works in the tomorrow nightly, but let me know if you see any issue

PR #1369, opened 2016-11-26
ENH: Add BrainExtractionTool extension

acsenrafilho:
Description:
This extension aims to extract the brain from the skull. This tool was
intensively tested with T1 MRI images, but the pipeline could be applied
to any other structural MRI images, namely T2 and PD. This module is an
adaptation of ROBEX brain extraction tool, given by Iglesias, J. E. et
al (2011). Robust Brain Extraction Across Datasets and Comparison With
Publicly Available Methods. IEEE Transactions on Medical Imaging, 30(9).
DOI:10.1109/TMI.2011.2138152

Contributors:
Antonio Carlos da S. Senra Filho (University of Sao Paulo)

### DIFF
--- a/BrainExtractionTool.s4ext
+++ b/BrainExtractionTool.s4ext
@@ -1,0 +1,13 @@
+build_subdirectory .
+category Segmentation
+contributors Antonio Carlos da S. Senra Filho (University of Sao Paulo)
+depends NA
+description This extension aims to extract the brain from the skull. This tool was intensively tested with T1 MRI images, but the pipeline could be applied to any other structural MRI images, namely T2 and PD. This module is an adaptation of ROBEX brain extraction tool, given by Iglesias, J. E. et al (2011). Robust Brain Extraction Across Datasets and Comparison With Publicly Available Methods. IEEE Transactions on Medical Imaging, 30(9). DOI:10.1109/TMI.2011.2138152
+enabled 1
+homepage http://slicer.org/slicerWiki/index.php/Documentation/Nightly/Extensions/BrainExtractionTool
+iconurl https://www.slicer.org/w/images/7/73/BrainExtractionTool-logo.png
+scm git
+scmrevision d58bec0189ca6d449d64c304cb3c46156a052fcb
+scmurl https://github.com/CSIM-Toolkits/BrainExtractionTool.git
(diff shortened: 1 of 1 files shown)


acsenrafilho:
Hello Slicers!

This extension aims to help the brain extraction procedure in Slicer, where here I made an adaptation of the original ROBEX brain extraction code (to fit a CLI module).

I hope that the compilation procedure goes fine (it has a ROBEX-data folder which I copy it in the HOME directory...but I do not know if this is a good way to do that)

Any comments is more than welcome! 

Cheers
Antonio

ihnorton:
There are several issues with this submission.

- Please do not call this "Brain Extraction Tool", to avoid confusion with the BET software from FSL. BET is widely-known and used, and has significant naming priority. "ROBEX Brain Extraction" or "ROBEX Skull Stripping" would be better.

- What is the license of this code? Most of the code appears to be the same as https://www.nitrc.org/projects/robex/ but without any license or attribution. Distributing that BSD-licensed code via Slicer extension build system is fine, but the original license and copyright must be preserved.

- The [qsort file](https://github.com/CSIM-Toolkits/BrainExtractionTool/blob/master/BrainExtractionTool/qsort.c) is from R. R is GPL licensed so using the code is fine, but taking it without respecting the license is not OK. That means this extension must be GPL licensed (again, fine, but the license must be preserved).

ihnorton:
Also: thanks for the submission! Having more skull-stripping/extraction tools for Slicer is great, we just need to keep in mind the necessary license aspects for redistributing code in order to avoid any problems for the Slicer community providing the build/distribution facility.

acsenrafilho:
Hi @ihnorton 
In fact, I did not paid attention to the "BET" naming. I will change it, no problem.

Also, I was a little confused about the license, as you pointed out. In fact, what I did here was only an adaptation of the original code (from https://www.nitrc.org/projects/robex/ ) in order to be used as a module in Slicer. But, what should I do to fix the correct license issues?

Can I put in the repository all the licenses that are listed in the ROBEX website (BSD, GPL and LGPL) and give a clear statement in the documentation page about it? 

Sorry for the initials problems.

lassoan:
There is no problems with license compatibility, just make sure you include all the original licenses in each file that contains code from somewhere else and add a summary of toolkits/types of licenses used in help/acknowledgment section. Thank you for your contribution!

jcfr:
Closing. This PR is superseded by #1371

PR #1387, opened 2017-01-16
Add ScatteredTransform extension

grandwork2:
Creates a BSpline transform from a displacement field defined at scattered points by using the Multi-level BSpline interpolation algorithm.

### DIFF
--- a/ScatteredTransform.s4ext
+++ b/ScatteredTransform.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl git://github.com/grandwork2/ScatteredTransform
-scmrevision d22c1aa
+scmrevision master
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files
--- a/ScatteredTransform.s4ext
+++ b/ScatteredTransform.s4ext
@@ -1,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl git://github.com/grandwork2/ScatteredTransform
+scmrevision d22c1aa
+
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends NA
+
+# Inner build directory (default is ".")
+build_subdirectory .
+
+# homepage
+homepage http://slicer.org/slicerWiki/index.php/Documentation/Nightly/Extensions/ScatteredTransform
+
+# Firstname1 Lastname1 ([SubOrg1, ]Org1), Firstname2 Lastname2 ([SubOrg2, ]Org2)
+# For example: Jane Roe (Superware), John Doe (Lab1, Nowhere), Joe Bloggs (Noware)
+contributors Grand Joldes (The University of Western Australia, ISML)
+
+# Match category in the xml description of the module (where it shows up in Modules menu)
+category Transforms
+
+# url to icon (png, size 128x128 pixels)
+iconurl https://raw.githubusercontent.com/grandwork2/ScatteredTransform/master/ScatteredTransform.png
+
+# Give people an idea what to expect from this code
+#  - Is it just a test or something you stand behind?
+status 
+
+# One line stating what the module does
+description Creates a BSpline transform from a displacement field defined at scattered points by using the Multi-level BSpline interpolation algorithm.
+
+# Space separated list of urls
+screenshoturls https://raw.githubusercontent.com/grandwork2/ScatteredTransform/master/Screenshot.png
+
+# 0 or 1: Define if the extension should be enabled after its installation.
+enabled 1


fedorov:
It appears that your extension depends on the Boost library: https://github.com/grandwork2/ScatteredTransform/blob/master/ScatteredTransform/ScatteredTransform.cxx#L35

Boost is not part of the core Slicer application. This code will not compile.

lassoan:
We used to have boost in an extension but the repository has been removed from gitorious and I couldn't find the code elsewhere. You have two options:

1. Keep using boost
You need to create an extension similar to SlicerOpenCV (https://www.slicer.org/wiki/Documentation/4.6/Extensions/SlicerOpenCV), which downloads and builds OpenCV library and makes it available for other extensions. An example is given here for building boost using CMake:
http://stackoverflow.com/questions/41644713/helloworld-downloading-building-and-linking-boost-using-cmake

2. Get rid of boost
I would recommend this, as boost is huge and no simple automated build solution is provided by the library developers.
You can already use ITK and VTK, which have shared pointers, transforms, etc. Most probably you can easily remove boost dependency.

grandwork2:
Option 1 is way too complicated for me to implement, due to the size of the boost library and its build system. 

Option 2 is also complicated, as the implementation of the MBA algorithm I am using is highly dependent on boost.

Nevertheless, I found another option. Because the extension only uses header files from boost (does not need the compiled libraries). I managed to include only those files in the extension (I selected them using the bcp tool from boost) and eliminate any external dependency to the library. Can you please try to compile it again?

lassoan:
After your changes, I could successfully build your extension. Its own test passed (after fixing TEMP path setting). These should be fixed in making the module usable in Slicer:
* I could not test the transform computation on actual data, as the module did not accepted markup fiducials already in the scene but expected me to select a file instead. This can be fixed by replacing     <file> by <point multiple="true">. See https://github.com/Slicer/Slicer/tree/master/Modules/CLI/FiducialRegistration as an example (it's very similar in the sense that it takes two sets of landmarks and computes a transform).
* Add set(TEMP "${CMAKE_BINARY_DIR}/Testing/Temporary") to ScatteredTransform\Testing\Cxx\CMakeLists.txt (without that I'm getting test failure because transform file cannot be written)
* There is no need to differentiate between ITK and Slicer coordinate systems, just specify <point multiple="true" coordinateSystem="ras"> or <point multiple="true" coordinateSystem="lps"> and Slicer will do the conversion for you
* Minor thing: Change the displayed module name from "ScatteredTransform" to "Scattered Transform"

grandwork2:
All done:
* I added support for fiducials input.
* Set up the temp folder* The selection between Slicer/ITK coordinate systems is included so that the generated transform file can be used outside Slicer (directly from ITK). I the end this is an CLI module and can be also used outside Slicer.
* Changed displayed name.

One question: how do I create the documentation page? I have access to Slicer Wiki, but how do I get the page started? 

    On Wednesday, 18 January 2017, 11:47, Andras Lasso <notifications@github.com> wrote:
 

 After your changes, I could successfully build your extension. Its own test passed (after fixing TEMP path setting). These should be fixed in making the module usable in Slicer:   
   - I could not test the transform computation on actual data, as the module did not accepted markup fiducials already in the scene but expected me to select a file instead. This can be fixed by replacing by . See https://github.com/Slicer/Slicer/tree/master/Modules/CLI/FiducialRegistration as an example (it's very similar in the sense that it takes two sets of landmarks and computes a transform).
   - Add set(TEMP "${CMAKE_BINARY_DIR}/Testing/Temporary") to ScatteredTransform\Testing\Cxx\CMakeLists.txt (without that I'm getting test failure because transform file cannot be written)
   - There is no need to differentiate between ITK and Slicer coordinate systems, just specify or and Slicer will do the conversion for you
   - Minor thing: Change the displayed module name from "ScatteredTransform" to "Scattered Transform"

lassoan:
Thank you. Could you push your changes to github so that we can test it?

Slicer (RAS) and ITK (LPS) coordinate system conversion is already provided by Slicer. It shouldn't be necessary to duplicate that mechanism. Have you tried to specify "ras" and "lps" coordinate system for the points element? Has neither option covered all your use cases?

Probably category should be Registration / Specialized, as that's where similar modules are located already.

grandwork2:
I changed the category and pushed the changes.Regarding the use cases: the main use for this module, from my point of view, is to transform images based on registration results obtained using finite element or meshless methods. Therefore, you get the initial and the displaced positions of the nodes in a file and want to use this info to transform the image. Using the finite element or meshless shape functions to do space interpolation is very time consuming; this module creates an alternative BSpline transform which is very cheap to evaluate at each voxel. 

Now, if I create a BSpline transform using this scenario and want to use it in Slicer, I need to transform the coordinates read from the files into Slicer's coordinate system (otherwise the transform created by the module will not work properly in Slicer). If I want to use the transform file directly from ITK (for example in another software, outside of Slicer, to find the displaced positions of other points of interest in the model), then I don't need any coordinate transformation. 

There may be a simpler way of making this coordinate transformation. I saw the FiducialRegistration module is doing it in a similar way. How do you specify "ras" and "lps" coordinate system for the points element?

 

    On Wednesday, 18 January 2017, 22:11, Andras Lasso <notifications@github.com> wrote:
 

 Thank you. Could you push your changes to github so that we can test it?Slicer (RAS) and ITK (LPS) coordinate system conversion is already provided by Slicer. It shouldn't be necessary to duplicate that mechanism. Have you tried to specify "ras" and "lps" coordinate system for the points element? Has neither option covered all your use cases?Probably category should be Registration / Specialized, as that's where similar modules are located already.—

lassoan:
Thank you , it is almost good now!

The xml file is corrupted, please change doc line to this:
<documentation-url>http://slicer.org/slicerWiki/index.php/Documentation/Nightly/Extensions/ScatteredTransform</documentation-url>

Remove transformCS field (in your module you can just always use ITK's convention, LPS coordinate system) and instead request points in LPS coordinate system:
<point multiple="true" coordinateSystem="lps">

Make the "Input files" section collapsed by default (set <parameters advanced="true">)

Remove bsplineTransformFile, from your module's viewpoint it's the same as bsplineTransform (it defines a filename), but in Slicer bsplineTransform creates a transform node selector bsplineTransformFile creates a file selector (the latter is not very usable).

Move all parameters for that you can provide a reasonable default (probably all of them) into a by-default-collapsed (advanced="true") section to make the module easier to use.

grandwork2:
I will implement these changes.
I have just one question regarding the transformCS field: how do you request points in LPS coordinate system? Do you mean by specifying coordinateSystem="lps" in the XML file for fiducials?
I see 2 scenarios that the module should be able to handle:1. You get the fiducials from Slicer - you may be able to request a change in the coordinate system - but the mechanism using XML config is static (if you request the points in LPS CS the resulting transform will work properly in ITK only but not in Slicer, if you request in RAS CS the transform will work in Slicer only). 
2. You get points from files (ITK convention), create a transform, then load those points (or an image in the same space as the points) in Slicer and apply the transform to them. This will not work properly if the transform is created using ITK coordinates (the resulting transform will not work correctly in Slicer). 

The indication of the coordinate system used should somehow be a transform attribute, but I don't know how this can be done. If you plan to use the generated transform in Slicer, it needs input in RAS CS; if you plan to use it in ITK, it needs input in LPS CS. 

Maybe the name and the description of the parameter does not reflect well this idea. I will try to find a better way of describing the parameter.

 


    On Thursday, 19 January 2017, 4:38, Andras Lasso <notifications@github.com> wrote:
 

 Thank you , it is almost good now!The xml file is corrupted, please change doc line to this:
http://slicer.org/slicerWiki/index.php/Documentation/Nightly/Extensions/ScatteredTransformRemove transformCS field (in your module you can just always use ITK's convention, LPS coordinate system) and instead request points in LPS coordinate system:
Make the "Input files" section collapsed by default (set )Remove bsplineTransformFile, from your module's viewpoint it's the same as bsplineTransform (it defines a filename), but in Slicer bsplineTransform creates a transform node selector bsplineTransformFile creates a file selector (the latter is not very usable).Move all parameters for that you can provide a reasonable default (probably all of them) into a by-default-collapsed (advanced="true") section to make the module easier to use.—

lassoan:
> Do you mean by specifying coordinateSystem="lps" in the XML file for fiducials?

Yes. See specification here https://www.slicer.org/wiki/Documentation/Nightly/Developers/SlicerExecutionModel and examples here https://github.com/Slicer/Slicer/blob/master/Modules/CLI/ExecutionModelTour/ExecutionModelTour.xml

I understand that you would like to use the module from the command-line for ITK images and transforms, and also from Slicer. That's why you need to specify in the XML that your module expects point coordinates in LPS. Slicer takes care of the rest (converts coordinates from its internal RAS representation to LPS when it calls modules that need coordinates in LPS). You don't need to specify RAS/LPS in a separate flag or implement conversion in your code.

grandwork2:
I have corrected the XML and organized the input parameters. 

Your proposal for handling inputs will not work.Consider I have 2 files with input points, F1 with points in the initial position and F2 with points in the displaced position. Both are in LPS CS. 
 I need to be able to generate 2 different transforms T_ITK and T_Slicer in order to support the following use-cases:Case 1: Read F1, F2 and T_Slicer in Slicer. Apply (in Slicer) T_Slicer to F1 and obtain F2.Case 2: Read F1, F2 and T_ITK in ITK. Apply (in ITK) T_ITK to F1 and obtain F2.
In these 2 use-cases T_ITK and T_Slicer have to be different! When the transforms are applied, T_Slicer will receive its input in RAS CS (because of the internal CS transformation performed by Slicer) and T_ITK will receive its input in LPS CS. 

If I create T_Slicer using inputs in LPS CS, the transform will not work properly in Slicer - therefore for this case I need to change the CS of the input points before creating the transform. The same is done in the FiducialRegistration module. When I create T_ITK I don't need to make any change to the input points.

A second point: the bsplineTransformFile is needed in order to save 1D and 2D transforms. Slicer crashes if I try to save such transforms in a Slicer node. I moved this in the Advanced Parameter section.
 

    On Thursday, 19 January 2017, 9:59, Andras Lasso <notifications@github.com> wrote:
 

 
Do you mean by specifying coordinateSystem="lps" in the XML file for fiducials?
Yes. See specification here https://www.slicer.org/wiki/Documentation/Nightly/Developers/SlicerExecutionModel and examples here https://github.com/Slicer/Slicer/blob/master/Modules/CLI/ExecutionModelTour/ExecutionModelTour.xmlI understand that you would like to use the module from the command-line for ITK images and transforms, and also from Slicer. That's why you need to specify in the XML that your module expects point coordinates in LPS. Slicer takes care of the rest (converts coordinates from its internal RAS representation to LPS when it calls modules that need coordinates in LPS). You don't need to specify RAS/LPS in a separate flag or implement conversion in your code.—

lassoan:
If you don't believe me that Slicer can emulate ITK's transforms, just test your module with coordinateSystem="lps" and keeping transformCS flag at "ITK". I've already tried this and it worked well for me, but maybe I missed something, so if you experience any problems in any scenario (using the executable standalone or in Slicer) then let me know and I'll look into it.

Slicer will only read/write/use 3D transforms, but it should not crash if it gets something else. I'll fix transform node so that it gracefully fails and logs a meaningful error message if it receives a 1D or 2D transform. Outside Slicer you can still use the executable to generate 1D/2D transforms and write them to file, so there is no need to define separate output transforms for general and Slicer use.

lassoan:
To make the test pass, remove
--transformCS Slicer
from from ScatteredTransform\Testing\Cxx\CMakeLists.txt

grandwork2:
I sorted the test.
I added some test data so you can test my use-case. In the Data\Input folder you can find 2 .csv files containing input points and 2 .vtp files containing little spheres (glyphs) created based on these points. 

If you run the extension using the default settings with inputs from the csv files (ignore first value !) and then load the .vtp files in Slicer, you can apply the resulting transform to "points" and these will be moved over "points_displaced". If you change the option (which I renamed to intendedUse) from Slicer to ITK, no coordinate transformation is performed, the transform is created in a totally different CS and does not work (but it will work in an ITK program, if the points are read in ITK and the transform applied in ITK).

Regarding the writing to a transform node - even if you prevent Slicer from crashing, I would like to be able to create 1D or 2D transforms from within Slicer. If I write them to a node and Slicer fails gracefully, I can't save the results. 

    On Thursday, 19 January 2017, 12:10, Andras Lasso <notifications@github.com> wrote:
 

 If you don't believe me that Slicer can emulate ITK's transforms, just test your module with coordinateSystem="lps" and keeping transformCS flag at "ITK". I've already tried this and it worked well for me, but maybe I missed something, so if you experience any problems in any scenario (using the executable standalone or in Slicer) then let me know and I'll look into it.Slicer will only read/write/use 3D transforms, but it should not crash if it gets something else. I'll fix transform node so that it gracefully fails and logs a meaningful error message if it receives a 1D or 2D transform. Outside Slicer you can still use the executable to generate 1D/2D transforms and write them to file, so there is no need to define separate output transforms for general and Slicer use.—

lassoan:
If you set coordinateSystem="lps" then Slicer will transform point coordinates from RAS to LPS before sending it to the CLI module. So, fcsv file that you save with the scene (that contains points coordinates in RAS) will be different from the fcsv file that your CLI module gets (that contains points coordinates in LPS).

Images are always stored in LPS coordinate system (Slicer transforms them to RAS when reading and transforms back to LPS when writing out), therefore those work well, too.

I'm not sure if coordinate system switch mechanism is implemented for surface meshes, but I can implemented that in Slicer to make CLI modules simpler (and not need to worry about RAS/LPS coordinate system transformations internally).

The main differences between <file> and <transform> elements that when you use them in Slicer:
1. <transform> creates a transform node selector widget for selecting the input/output node on the GUI
2. Slicer deletes the transform file that the CLI creates after the <transform> is read from the file.

Behavior 1. is always preferable in Slicer (and makes no difference when you run the module without Slicer GUI). However, if you want Slicer to create a file that is not saved with the scene, not loaded into the scene, and not deleted after execution is completed, then additional file output may be useful.

lassoan:
ScatteredTransform\ScatteredTransform.cxx
int ret;
uninitialized - returns random results in some cases.

grandwork2:
I already assumed all the input data (either from fiducials or from files) is in LPS coordinate system. It seems this assumption was wrong for fiducials, so I have to set coordinateSystem="lps" when getting fiducial input. I did this and now all input is in LPS CS.
I tested this with fiducials input, and indeed it worked without any changes to the coordinates.I then tested with .vtp input (the data I specified in a previous email) and it does not work without coordinate transformation of the input data.
It seems to me that Slicer uses different CS when applying transforms to different data types! It uses LPS when handling fiducials but RAS when handling models (and images, I suppose).
It would be great if only LPS CS would be used, then no coordinate transformations would be necessary when generating the transform, but it seems right now this is not the case. Can you please check.

 

    On Thursday, 19 January 2017, 13:59, Andras Lasso <notifications@github.com> wrote:
 

 If you set coordinateSystem="lps" then Slicer will transform point coordinates from RAS to LPS before sending it to the CLI module. So, fcsv file that you save with the scene (that contains points coordinates in RAS) will be different from the fcsv file that your CLI module gets (that contains points coordinates in LPS).Images are always stored in LPS coordinate system (Slicer transforms them to RAS when reading and transforms back to LPS when writing out), therefore those work well, too.I'm not sure if coordinate system switch mechanism is implemented for surface meshes, but I can implemented that in Slicer to make CLI modules simpler (and not need to worry about RAS/LPS coordinate system transformations internally).The main differences between and elements that when you use them in Slicer:   
   -  creates a transform node selector widget for selecting the input/output node on the GUI
   - Slicer deletes the transform file that the CLI creates after the is read from the file.
Behavior 1. is always preferable in Slicer (and makes no difference when you run the module without Slicer GUI). However, if you want Slicer to create a file that is not saved with the scene, not loaded into the scene, and not deleted after execution is completed, then additional file output may be useful.—

lassoan:
I would suggest to always use the same coordinate system.

Hardcoded LPS: image, transform
Configurable LPS/RAS: point, pointfile
Hardcoded RAS: geometry

So, the simplest is to always use LPS. Only geometry does not support LPS right now, but I can implement coordinateSystem support in 1-2 days. Your module doesn't have geometry input/output, so not impacted directly anyway.

I've fixed the crash with 1D and 2D transforms and will commit later today.

By the way, model nodes in Slicer will soon handle volumetric meshes, too. So, for exanple you'll be able to visualize FEM meshes directly in Slicer, not just the outer surface. The pull request has been sent already and will be merged this week or next week.

lassoan:
The important thing is that the module works well and it's easy to use in Slicer (most users will not open the Advanced section, so you can put anything there, even if it's not necessary or may confure some people).

I would at least change "ITK" and "Slicer" names by "LPS" and "RAS" coordinate system (something like pointFileCoordinateSystem: "LPS" or "RAS"), but it's up to you.

Just one more thing to decide before merging: are you sure you want to specify a git tag in the s4ext file? If you specify a git tag then you can make changes in your repository without worrying that you break something (Slicer will not use the latest master version), but you have to send a pull request whenever you want your Slicer module to update.
If you specify "master" as scmrevision then Slicer build machines will get the latest master version of your extension and use that in the nightly builds. We've found that this makes extension updates much faster and simpler.

lassoan:
@grandwork2 Thank you for your contribution, I've added your extension to the index! Check out the dashboard tomorrow to see if the extension was correctly built on all platforms.

grandwork2:
Thank you. I have just one more problem to solve. I added support for output to be displayed directly in Slicer. This should make the user experience much simpler, especially when input is incorrect. The problem is: Slicer does not display the values from the returnParameterFile unless the module returns 0. How should I handle this? Should I return 0 in case of an error or you can modify the behavior of Slicer?

lassoan:
There are two options:

A. Print the error message on stdout and return with non-zero value. User will see that there is an error and he can open the log window. This is what most CLI modules do.

B. Add a string output field to the XML file and always return with 0. This status field would tell the user if execution was successful or if there was an error.

I would suggest option A. Probably we could make it easier for the user to look at the output and error messages by adding a section on the CLI module GUI that shows the last output/error.

grandwork2:
Is there a specific reason why Slicer does not display the values from the returnParameterFile when the module returns an error?

lassoan:
I guess this behavior was chosen to avoid inconsistencies caused by partial updates.

grandwork2:
I checked the dashboard. MSBuild completed fine, but Build Name: 25643-ScatteredTransform-git51cef49-clang++-64bits-Qt4.8-Release fails wit the following error:

/.../ScatteredTransform/ScatteredTransform/ScatteredTransform.cxx:37:10: fatal error: 'boost\bind.hpp' file not found

#include "boost\bind.hpp"

Seems to be related to the way clang compiler treats include paths. 

Any idea what's the best way to fix it? I don't have access to a clang compiler right now (or a mac computer).

grandwork2:
I think I found the error - the use of \ in include. Will change it.

lassoan:
@grandwork2 FYI, I've implemented display of CLI module output in the module widget - see https://github.com/Slicer/Slicer/pull/650. So, if there is an error in your module (user didn't specify something, etc) then just log a message on stderr and return with non-zero - users will be able to see that very easily.

grandwork2:
Excellent. I will modify the extension.

grandwork2:
The extension now compiles and passes the test on all platforms. I still have to complete the documentation page.
Is there anything else I have to do in order to have it included in Slicer?

lassoan:
It shows up in the Extension manager, in Transforms category.

Probably the issue is that category is not listed on the main page. Please change the extension's category to Registration.

PR #1394, opened 2017-02-16
Update NeedleFinder

CoolZiyang:
We add some functionality as well as update user interface so that user could switch back and forth among needles being segmented. Here is the comparison links and the summary.

https://github.com/gpernelle/NeedleFinder/compare/6cf994c5e5f52d3c9bd6d6c97b737e6ed1a4b73a...b1038b5b1ed04bd73c61f69a815fbcac93bb713d


fedorov:
Do you plan to follow up with updating master branch as well?

CoolZiyang:
Yes, I'm doing it now. When can we have the extension distributed on Slicer 4.6?

PR #1397, opened 2017-03-28
Update SlicerPathology.s4ext

ebremer:


lassoan:
@ebremer @tdiprima I would recommend to check out the new Segment editor module. It supports overlapping labels, contour+fill visualization, lots of new and vastly improved effects, and the Segment editor will be kept improved (while the old Editor module is in maintenance mode). To try: use the latest nightly build; use the Extension wizard module and choose scriptedsegmenteditoreffect module type to create a new effect. Contact me if you have any questions.

PR #1409, opened 2017-04-24
update DSCMRI Module info

AndrewBeers:
I will be maintaining this module for the future - would like to be reflected in the download menu.


fedorov:
I think it also makes sense to add Jayashree name

fedorov:
Thanks!

PR #1427, opened 2017-06-09
update Multi-LevelRegistration

emily-hammond:


lassoan:
Thank you for your submission. The module implements useful functionality and nicely documented. I would have just two suggestions that would be nice to be addressed, before integrating:

* Project and repository name: I would suggest to rename Multi-LevelRegistration to MultiLevelRegistration (project name in CMake files) and the GitHub repository to SlicerMultiLevelRegistration (the Slicer prefix indicates that it is a Slicer extension)
* No tests are implemented: It would be important to have at least a very basic test. It would be indicate that your module is functional on all platforms, even as Slicer core and libraries are changing. It also serves as a specific example that shows how your module works. There are lots of CLI testing examples that you can follow: https://github.com/Slicer/Slicer/tree/master/Modules/CLI.

emily-hammond:
@lassoan I have implemented the suggested naming schema. I have also put in some testing modeled after the PETStandardUptakeValueComputation module. I'm not familiar enough with CMake to understand what testing to implement (especially since it is a registration algorithm) anything more specific. Let me know if I need to do a new pull request or make any more changes.

lassoan:
Thanks for the update. The test should run the registration on a sample data set and check the output. 

There are lots of registration test for BRAINSFit registration module that you can use as an example: https://github.com/BRAINSia/BRAINSTools/blob/master/BRAINSFit/TestSuite/CMakeLists.txt

You can find simpler examples (that don't use external data, just files located in your repository) among Slicer CLI module tests, for example this one: https://github.com/Slicer/Slicer/blob/master/Modules/CLI/BlobDetection/Testing/CMakeLists.txt

jcfr:
I am integrating this PR.

suggestion posted by @lassoan have been added to https://github.com/emily-hammond/SlicerMultiLevelRegistration/issues/1

PR #1433, opened 2017-07-10
ENH: First push of BoneTextureExtension

jbvimort:
This extension contains several modules that can be used to compute feature maps of N-Dimensional images using well-known texture analysis methods, it is based on the ITK remote module itkTextureFeatures (https://github.com/InsightSoftwareConsortium/ITKTextureFeatures / http://www.insight-journal.org/browse/publication/985)

Key Features:
	-8 cooccurrence textural features: energy, entropy, correlation, inertia, cluster Shade...
	-10 run length textural features: run length emphasis, grey level non uniformity, run length non uniformity, low grey level long run emphasis...
	-Input configurable parameters: locality of the texture, offset directions for co-occurrence and run length computation, the number of bins for the intensity histograms, and the intensity range or the range of run lengths.

This extension is composed of:
	 - 5 CLIs:
		  -- 2 that compute textural features for the whole image: computeGLCMFeatures and computGLRLMFeatures
		  -- 2 that compute textural features map: computeGLCMFeatureMaps and computGLRLMFeatureMaps
		  -- the last CLI is used in order to divide a dwi image containing n components in n separated volume images
	 - 2 scripted modules:
		  -- BoneTexture: this module provides to the user a dynamic interface able to run the previous CLIs, modify the parameters, and observe the results
		  -- BoneTextureSerializer: this module provides an efficient way to apply the previous CLIs to several cases contained in the same input folder

For more information the extension documentation is available here: https://www.slicer.org/wiki/Documentation/Nightly/Extensions/BoneTextureExtension

### DIFF
--- a/BoneTextureExtension.s4ext
+++ b/BoneTextureExtension.s4ext
@@ -1,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl https://github.com/Kitware/BoneTextureExtension.git
+scmrevision nightly
+
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends NA
+
+# Inner build directory (default is ".")
+build_subdirectory BoneTextureExtension-build
+
+# homepage
+homepage https://www.slicer.org/wiki/Documentation/Nightly/Extensions/BoneTextureExtension
+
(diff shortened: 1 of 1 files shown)


fedorov:
Are the filters from itkTextureFeatures remote module already available from the ITK that is used in Slicer?

It seems from the comment in cmake file that the version of ITK used in Slicer predates the Insight Journal paper you referenced: https://github.com/Slicer/Slicer/blob/master/SuperBuild/External_ITKv4.cmake#L44.

fedorov:
Oh, nevermind - I see you build it in superbuild! sorry

PR #1435, opened 2017-07-19
Add Segmentation Wizard

AndrewBeers:
Hello!

This pull request is to add my new module, 3D Model Segmentation, to 3D Slicer. It is a multi-step module modeled after Andrey Fedorov's ChangeTracker. It uses the Volume Clip with Model module to make a quick and easy segmentation-and-thresholding tool for high resolution images. It includes utilities for creating subtraction maps and registering images.

I will have to change some of the attributions and acknowledgements before this module is ready to be integrated into 3D Slicer, but I should have that information in the next day or two. I wanted to get the process started to see if I am missing anything right off the bat. You can see the code for the module itself at: https://github.com/QTIM-Lab/3D_Model_Segmentation.

I have added a unit test, again in the mold of ChangeTracker. The unit test passed on Slicer 4.6 and the nightly Slicer from 6/30/17. However, I haven't found any documentation on how such unit tests are integrated into Slicer continuous integration (if they are), so I've mostly just tried to stay as close to the ChangeTracker model as possible.

I've created a documentation page at https://www.slicer.org/wiki/Documentation/Nightly/Extensions/3D_Model_Segmentation. I can copy this page to the 4.6 version of the page if necessary.

I've included information in the .s4ext file as best I understand from https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/Contribute_Extension_Description_File and https://www.slicer.org/wiki/Documentation/Nightly/Developers/Extensions/DescriptionFile. However, I'm not sure if there is way to test how the extension would "look" once uploaded into Slicer, and would love to read any existing documentation on how to build Python modules in Slicer so I could fix any problems before this extension is accepted.

A long-term plan is to modify this module to include built-in automatic segmentation algorithms, island removal tools, support for multi-label segmentations, and output to DICOM-SEG format and integration with the current Slicer Segmentations module. However, this module has proved useful to quite a few users already, so I wanted to release it now. I can squash all of my commits into one commit once ready to submit.

All the best,
Andrew

### DIFF
--- a/SegmentationWizard.s4ext
+++ b/SegmentationWizard.s4ext
@@ -1,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl git://github.com/QTIM-Lab/SlicerSegmentationWizard.git
+scmrevision master
+
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends NA
+
+# Inner build directory (default is ".")
+build_subdirectory .
+
+# homepage
+homepage https://www.slicer.org/wiki/Documentation/Nightly/Extensions/SegmentationWizard
+
+# Firstname1 Lastname1 ([SubOrg1, ]Org1), Firstname2 Lastname2 ([SubOrg2, ]Org2)
+# For example: Jane Roe (Superware), John Doe (Lab1, Nowhere), Joe Bloggs (Noware)
+contributors Andrew Beers (Massachusetts General Hospital)
+
+# Match category in the xml description of the module (where it shows up in Modules menu)
+category Segmentation
+
+# url to icon (png, size 128x128 pixels)
+iconurl https://raw.githubusercontent.com/QTIM-Lab/SlicerSegmentationWizard/master/SegmentationWizard/Resources/Icons/SegmentationWizard.png
+
+# Give people an idea what to expect from this code
+#  - Is it just a test or something you stand behind?
+status Beta 
+
+# One line stating what the module does
+This is a 3D model-based segmentation tool for 3D Slicer. It includes utilities for calculating subtraction maps and thresholding intensities.
+
+# Space separated list of urls
+screenshoturls https://raw.githubusercontent.com/QTIM-Lab/SlicerSegmentationWizard/master/SegmentationWizard.png https://www.slicer.org/w/images/c/c9/3D_Model_Segmentation_Step_3B.PNG https://www.slicer.org/w/images/9/9a/3D_Model_Segmentation_Step_4B.PNG https://www.slicer.org/w/images/8/8b/3D_Model_Segmentation_Step_5B.PNG
+
+# 0 or 1: Define if the extension should be enabled after its installation.
+enabled 1


lassoan:
Thank you for your contribution, your module looks very nice and it is a good example for implementing a multi-step segmentation workflow. Documentation and video tutorial are also clear and thorough.

I think the module could be offered in the Slicer 4.6 app store without changes (just changing the extension name something more specific). However, in Slicer 4.7 (nightly) the module should use Segment Editor module, instead of the legacy Editor module.

Segment Editor contains improved effects of the old editor, powerful new effects (such as 3D scissors, logical operators, interpolate between slices), and many new fundamental features (real-time 3D display, support for overlapping segments, masking, etc). SegmentEditorExtraEffects extension includes the "Surface cut" effect, which is essentially makes markups to model functionality available as an editor effect. This effect, combined with built-in threshold masking capability of the segment editor provides very similar functionality as the second half of steps in your module.

It would be still good to have your module in the 4.7 app store, as  there is no background subtraction functionality in Segment editor. This is important, because powerful region growing effects in Segment Editor, such as "Grow from seeds" and "Watershed" often gives better results than surface cut+thresholding with comparable user effort. Also, the wizard-like interface that you implemented may be easier to use for multi-step workflows compared to just showing all available tools and let the user figure out what tools he should use in what order.

Overall, I would recommend to:

1. Find a more specific name for your extension. Maybe something like ContrastImageSegmentation, ImageDifferenceSegmentation, ImageSubtractionSegmentation? It is also recommended to start the github repository name with "Slicer" to make it clear that the code is to be used in 3D Slicer. So, the repository name would be Slicer+ChosenExtensionName.

2. To be added to Slicer 4.7 app store, update the module to use Segment editor to not confuse users, and use the "Surface cut" effect to reduce code duplication.

fedorov:
I agree the use of Segment Editor is a good idea, and is desirable.

I disagree the use of Segment Editor over legacy Editor can be a mandatory condition for including a new extension in the distribution. 

The authors of the extension should have the full right to decide what they want to implement in their extension, as long as their extension meets the established rules. There is no community guideline that mandates the use of Segment Editor over Editor. Legacy Editor is still in 4.7, and there are many modules that still use legacy Editor.

lassoan:
> I disagree the use of Segment Editor over legacy Editor can be a mandatory condition

Of course. Nobody talked about any mandatory condition especially not as a general rule. I've spent some time with testing the extension, checking out documentation and tutorial, and I am very familiar with what is available in Slicer - so I feel quite confident that my recommendations serve the best interests of both users and the developers of this extension.

fedorov:
@lassoan - Great, I am glad I misinterpreted your earlier post!

pieper:
Excellent extension and valuable discussion - 👍 

I agree that the name could be a bit more descriptive and some example images on the wiki would be helpful so that people can get a quick idea of the use cases and expected results.

AndrewBeers:
Hello everyone --

Thanks for all of your comments. I agree that integrating the Segmentation tool would be useful, and it is definitely one of my priorities. I wanted to stick to the Editor tool for now because that is what my lab and myself is used to, but in the future it can be an additional option.

Steve -- I'll get working on adding a gallery with some additional screenshots.

As for the name, I agree that 3D Model Segmentation is not very evocative, but I'm hesitant to peg it to something more specific like ImageSubtractionSegmentation. We have plans to update this module over time, adding utilities for implementing automatic segmentation algorithms inside the user's 3d model "bubble". I've also found that the module is very useful for segmenting any number of tissues and situations -- most recently I've used it to quickly create lung masks in a lung DCE volume where automatic segmentation algorithms would not work. In light of these things, I'd like to keep the module name a bit wider open. Going in the vein of Slicer's Wizards tab, how about 3D Segmentation Wizard? Forefronting the "3D" part of the name is important to me, as the module was developed in the first place to make it easier to segment high resolution 3D volumes like MPRAGE.

I've added the acknowledgments I need, so apart from the name change, this module is ready to be integrated.

Thanks for are your help again --
Andrew

fedorov:
@AndrewBeers what kind of input DICOM series does this module expect? Do you typically have pre- and post-contrast volumes as separate series in the DICOM study?

lassoan:
"Segmentation Wizard" could be a good name if you plan to add several additional segmentation wizard workflows (which would be great!).

I'm not sure adding "3D" would be a good idea, as in 3D Slicer everything is in 3D and actually you cannot segment in 2D (if you load a 2D image, it is loaded as a single-slice 3D volume). "3D" prefix is not just unnecessary, but may confuse novice users, who might think that some other extensions (that don't have "3D" in their name) are not working as well in 3D.

If you work in such a generic namespace as "Segmentation wizard" then others may need to include their segmentation wizard modules in your extension. They will send pull requests, they have to be added to the list of contributors, etc., which may put a little additional burden on you and branding of your extension will not that be trivial (you cannot simply claim that you develop this extension, just something like that you are the lead developer or something similar). If this would be a problem then you might make your extension more specific by adding one more word before or after SegmentationWizard.

lassoan:
> I wanted to stick to the Editor tool for now because that is what my lab and myself is used to

You should really switch to the Segment Editor. It takes Slicer's editing capabilities from a lower-mid range to top-class (best in class in many aspects). Note that point-based surface generation is already available as a Segment Editor effect (Surface cut effect, it shows up in Segment Editor if you install SegmentEditorExtraEffects extension).

AndrewBeers:
@fedorov As configured currently, this module only expects MRML nodes, and doesn't contain any information about DICOM. I don't intend to limit it to any particular sort of sequence, modality, or series, but in our own case, pre- and post-contrast volumes have come as separate series in our studies. It makes sense, however, to make the subtraction map capabilities of this module extend to MultiVolumes -- that would be an interesting and fairly unique Slicer use-case.

@lassoan I decided to go with "Segmentation Wizard". It is a goal within our lab to build on this framework and add additional workflows, and I don't mind others making pull requests to the main module -- it probably help me learn some new way to take advantage of the Slicer API. I also agree about the Segment Editor -- it looks great, and I've already started using it! I only meant that this module had been originally coded with the Editor tool, and that I'd rather submit now to have something out there and then update soon with the Segment Editor.

@pieper I've added a gallery to https://www.slicer.org/wiki/Documentation/Nightly/Extensions/SegmentationWizard

I've squashed my commits into one, and added acknowledgements, so this should be ready to be uploaded on my end.

pieper:
Sounds great to me - thanks Andrew!

fedorov:
> in our own case, pre- and post-contrast volumes have come as separate series in our studies

Ok, great!

lassoan:
Thanks for the updates. It is almost ready to merge, I've just noticed that you haven't updated the repository name. Currently it is "3D_Model_Segmentation", while it should be "SlicerSegmentationWizard" (extension name prefixed with "Slicer", unless the extension name already starts with Slicer). You can easily rename a repository in github.

AndrewBeers:
Sounds good - I've renamed the repository and updated links as necessary.

lassoan:
👍  Thank you for your contribution. Check out the [Slicer dashboard](http://slicer.cdash.org/index.php?project=Slicer4) tomorrow to make sure there are no errors on any platforms.

AndrewBeers:
@lassoan Thanks! I did check the dashboard, and it looks like there was an error that I had added at the last moment having to do with the module name change. I fixed this error and pushed this on the Github repository. Do I need to resubmit in some way, or should the changed code be reflected in the next nightly build?

lassoan:
You don't need to do anything else. Your changes will be included automatically in the nightly build tomorrow.

PR #1440, opened 2017-07-25
UPDATE: Anomalous Filters Extension

acsenrafilho:
Further bug fixes were added in this version in order to correct iOS building process.
Also, the git hash was changed to master.


acsenrafilho:
Thanks @lassoan for this git hash tip!!

PR #1441, opened 2017-07-26
ENH: updated PETDICOMExtension revision to be used in ExtensionManager

che85:


fedorov:
@chribaue should we change this to track master?

chribaue:
@fedorov yes, we can

PR #1454, opened 2017-08-24
ENH: Remove Eigen "build time" extension

jcfr:
Impact:

No extension explicitly depends on Eigen

Only the PortPlacement extension depends on Eigen  and will be updated
to have a superbuild structure.

Motivation:

In its current form, the Eigen build tree is not usable:

(1) the "install" target is expected to be built

(2) the Eigen3Config.cmake references path that are expected to be set
using -DCMAKE_INSTALL_PREFIX::PATH=/path/to/install


jcfr:
Closing. It turns out that the latest version of Eigen works as expected.

PR #1460, opened 2017-08-26
ENH: Update Portplacement to fix build error and add Qt5 support

jcfr:


jcfr:
Thanks @giogadi for integrating https://github.com/giogadi/PortPlacement/pull/3

PR #1462, opened 2017-08-30
Added SlicerAutoscroll Extension

moselhy:


lassoan:
Thank you very much for your contribution. This is a nice extension, with sufficient documentation, A few comments/suggestions:
* For me it is not obvious that in what cases you want to use this tool (I always use the arrow buttons to easily rock a 1-2 slices back-and forth). Could you upload a video somewhere that shows how you use this tool during segmentation?
* The extension name is nice in the sense that it is quite unique, but I'm not sure if users can guess from the name what the extension might do. Instead of "cartooning", something like "slice rocker" or "slice auto-scroll" would be more descriptive. To test if a name is good enough, ask a few people around you to try to guess based on the name what the extension may do.
* Instead of the current continuous transition, it might be better to move slice offset by the slice thickness (jump to previous/next slice the same way as the arrow buttons do).
* Extension description should mention that this is a segment editor effect.
* Add the segment editor icon to the extension's icon to make it easier for users to recognize that this is a Segment editor effect. You can use this file as a template https://github.com/lassoan/SlicerSegmentEditorExtraEffects/blob/master/SegmentEditorExtraEffects.xcf

moselhy:
Great suggestions, thank you!


> I always use the arrow buttons to easily rock a 1-2 slices back-and forth

The extension is good for segmenting organs that have partial volume effects such as multiple placentas or when it is hard to see the borders of the organs. It is better than manually rocking back and forth with the arrow keys for two reasons. 
1) You don't have to click the slice view widget to be able to use the hotkey
2) when you press the hotkey again (Alt+C) it goes back to the slice you were on so you can segment it while keeping a vivid picture of the cartoon in your short term memory.



> ask a few people around you to try to guess based on the name what the extension may do

my lab uses the term 'cartooning' quite often

> jump to previous/next slice the same way as the arrow buttons do

Done in commit https://github.com/moselhy/SlicerCartooner/commit/8626a65730c01b1da43e55fe54d60b166b48b136



> Extension description should mention that this is a segment editor effect

Done in commit https://github.com/moselhy/SlicerCartooner/commit/30e524ff9d5db21447f15a5e496faec45ba8ed7a



> segment editor icon to the extension's icon

Done in commit https://github.com/moselhy/SlicerCartooner/commit/1c8a316e9ca6913b837c3af4ec8d5318fb4cf0d4

lassoan:
> my lab uses the term 'cartooning' quite often

As the name is very specific, I don't mind if you choose to keep it, but I suspect most users would have no idea what you mean by it. I've never heard this term used in this context.

Do you know anybody using this term outside your lab?
Can you find any websites, books, anything where this term is used?
Do you know why it is called "cartooning"?

pieper:
I'll jump in and say I had no idea what 'cartooning' meant in this context.  Looking at wiki page didn't really help much either.  Now that I've read this discussion I understand what it is and it sounds valuable, but I would also vote for a more descriptive name.

moselhy:
Renamed to SlicerAutoscroll in https://github.com/moselhy/SlicerAutoscroll/commit/4d864ab31aba0498432449f80f39e94a25669f20

lassoan:
Thanks for all the fixes, the extension is added now to the index.

I just have one suggestion: change helpText(self) method to return a string that starts with a short, one-line description and add have the remaining of the help text after the line break that. The full text is only visible if the user clicks on "Show details". You may also use html formatting, for example:

```
"""<html>Add uniform intensity region to selected segment<br>.
<p><ul style="margin: 0">
<li><b>Mouse move:</b> current background voxel is used to find a closed path that
follows the same intensity value back to the starting point within the current slice.</li>
<li><b>Left-click:</b> add the previewed region to the current segment.</li>
</ul><p></html>""" 
```

moselhy:
Done, thanks

PR #1468, opened 2017-10-09
EHN: Add ZFrameCalibration extension. 

leochan2009:
Description: This cli module takes a ZFrame MRI image as input, calculates the ZFrame transformation matrix in the RAS coordinates

### DIFF
--- a/ZFrameCalibration.s4ext
+++ b/ZFrameCalibration.s4ext
@@ -1,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl https://github.com/leochan2009/ZFrameCalibration.git
+scmrevision 8c316db
+
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends NA
+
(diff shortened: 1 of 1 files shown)


lassoan:
Thanks for your contribution. To be able to publish an extension in the extension index, a minimal amount of documentation is needed that would allow users to find out what the extension is for, interested users should have information about how they can try it (see this extension for an example: https://github.com/moselhy/SlicerSequenceRegistration, probably for you a simple tutorial would suffice).

A few specific suggestions:
- [ ] In https://github.com/leochan2009/ZFrameCalibration/blob/master/CMakeLists.txt, add valid images for icons and screenshots
- [ ] In https://github.com/leochan2009/ZFrameCalibration/blob/master/README.md, add one-paragraph summary, sample screenshots, example data set (you can upload to MIDAS, if you are not sure how to do it I can help), and basic tutorial about how a user can try this module.
-[ ] Consider using a more accurate module name: in itself, this module performs a registration operation, that an application may use for calibration. It would be more accurate to be called ZFrameRegistration.
- [ ] Extension **repository** name should start with `Slicer` (SlicerZFrameRegistration, the extension name would be still just ZFrameRegistration).

leochan2009:
Thanks Andras,

I will add the missing component soon.

Best,
Longquan

PR #1479, opened 2017-10-16
ENH: Remove FinslerTractography extension

jcfr:
From @petersv2 (Peter Savadjiev):

>
> After speaking with @ljod (Lauren O'Donnell), we decided it’s best to
> abandon this extension. It is based on old code that is now obsolete.
>
> We have a new and improved version of the FInsler code, located here:
> https://github.com/tomdelahaije/fcm but it is a stand-alone program,
> not a Slicer extension.  If we have the resources one day, we’ll try
> to create a new extension based on this new code.
>


ljod:
Thank you!

jcfr:
@petersv2 Would it be possible to update https://www.nitrc.org/projects/finslertract/ to mention the extension is not available anymore

jcfr:
> https://github.com/tomdelahaije/fcm but it is a stand-alone program,
> not a Slicer extension

@ljod  @petersv2 Worth noting that the *new* code uses the Slicer execution model, meaning it would be fairly trivial to built it as an extension ... if you have some funding it could be done be the Slicer team at kitware in few hours.

ljod:
Thanks for the information. Please let me know what that would cost.

PR #1488, opened 2017-10-21
Update Anomalous Filters Extension

acsenrafilho:
A new imaging modality was added to the 4.8 branch: EPI filtering.


jcfr:
Thanks for the PR. Superseded by #1490 

Moving forward, I would suggest to use the extension wizard to submit update. It has simple as:

```
cd AnomalousFiltersExtension
/path/to/Slicer-build/Slicer-build/bin/slicerExtensionWizard --target 4.8 --contribute
```

PR #1508, opened 2018-01-19
Added OpenIGTLinkIF extension. 

leochan2009:
The module was in the core,now it is moved to extension. The module is used to transfer data (image, transform, video,etc.) with other platform.

### DIFF
--- a/OpenIGTLinkIF.s4ext
+++ b/OpenIGTLinkIF.s4ext
@@ -1,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl git://github.com/openigtlink/SlicerOpenIGTLink.git
+scmrevision fd5ac89df2243ef4da13f4f4843e4d79ebf787d6
+
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends NA
+
+# Inner build directory (default is ".")
+build_subdirectory inner-build
+
+# homepage
+homepage https://github.com/openigtlink/SlicerOpenIGTLink
+
+# Firstname1 Lastname1 ([SubOrg1, ]Org1), Firstname2 Lastname2 ([SubOrg2, ]Org2)
+# For example: Jane Roe (Superware), John Doe (Lab1, Nowhere), Joe Bloggs (Noware)
+contributors Junichi Tokuda (SPL), Longquan Chen (SPL)
+
+# Match category in the xml description of the module (where it shows up in Modules menu)
+category IGT
+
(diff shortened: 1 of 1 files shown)


OpenIGTLinkIF.s4ext
```diff
@@ -0,0 +1,44 @@
...
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends NA
+
+# Inner build directory (default is ".")
+build_subdirectory inner-build
+
+# homepage
+homepage https://github.com/openigtlink/OpenIGTLinkIF
```
lassoan: Should it be https://github.com/openigtlink/SlicerOpenIGTLink instead?

OpenIGTLinkIF.s4ext
```diff
@@ -0,0 +1,44 @@
...
+
+# url to icon (png, size 128x128 pixels)
+iconurl https://github.com/openigtlink/SlicerOpenIGTLink/blob/master/OpenIGTLinkIF.png
+
+# Give people an idea what to expect from this code
+#  - Is it just a test or something you stand behind?
+status 
+
+# One line stating what the module does
+description This extension faciliates the communication between 3D Slicer and other platform via OpenIGTLink.
```
lassoan: Could be tuned a bit: This extension adds OpenIGTLink communication interface to 3D Slicer to exchange real-time imaging, tracking, and other data.

leochan2009: Sounds better:-)

leochan2009:
@lassoan ,
Just modified according to your suggestions.

Best,
Longquan

lassoan:
Thank you, it looks good. Just one more thing to consider: If you specify a git hash instead of a branch name then you have to submit a pull request here each time you update the extension. It is usually much more efficient to specify a branch name instead and then the factory machines will just get the latest version from that branch every night. Usually we use the branch names matching the corresponding branch names in Slicer or ExtensionsIndex repository (master and master-48 or 4.8).

leochan2009:
Thanks for your suggestion.
yes, it make sense to use the master branch now, as it is still under development. 
We probably going to change the extension in the next days.

lassoan:
I've merged it so that we have a chance to see any build errors on any platforms on the nightly dashboard.

leochan2009:
great!

jcfr:
I suggest to rename the file to `SlicerOpenIGTLink.s4ext`. From the build system perspective, this is what defined the name of the extension needed to manage dependency between extension.

Also, waiting the extension is proven to work, I am updating [Slicer/CMakeLists.txt](https://github.com/Slicer/Slicer/blob/926acc5a6660bc532290bf53e1aac381d1aaf2ec/CMakeLists.txt#L288-L289) to disabling the `Slicer_USE_OpenIGTLink` option. Failing to do so would prevent the extension from building successfully because of duplicated targets.

PR #1519, opened 2018-02-09
EHN: extension description file renamed

leochan2009:

PR #1520, opened 2018-02-09
EHN: rename the OpenIGTLinkIF.s4ext to SlicerOpenIGTLink.s4ext

leochan2009:


leochan2009:
#1519 
@lassoan @fedorov

lassoan:
Do we want to keep the IF at the end of the name? I think it would be more appropriate to just call the extension SlicerOpenIGTLink. It would contain only OpenIGTLinkIF module now, but we might add more modules in the future.

leochan2009:
will do so!

leochan2009:
@lassoan SlicerIGT is depending on SlicerOpenIGTLink, so this pull request need to be merged to make slicerigt build succeed.

Best
longquan

PR #1522, opened 2018-02-16
Adding SlicerVASST extension

adamrankin:


cpinter:
Yaay!

adamrankin:
It's mostly a lot of VTK classes that have slowly accumulated at the VASST lab over the years. I haven't tested any of them.

The one algo module is a point-to-line registration algorithm by Elvis Chen, which I have tested (and use) regularly.

adamrankin:
Ah, I thought the slicer wiki page was autogenerated from the extension index!

Unfortunately I don't have the bandwidth to maintain the extension aside from fixing build errors. Hopefully when I transition into part-time work it will be part of my duties. We'll see.

lassoan:
You can just change the homepage url to the github project url and add a Readme.md file with some minimal info. There should be at least as much information that people could figure out what these modules can do - such as a screenshot and a paragraph for each module.

lassoan:
There are screenshot URLs are invalid, too. I think "IGT" would be more appropriate as category (now it is Registration).

adamrankin:
There are no screenshots. Can the entry just be removed?

lassoan:
Users need some information about the extension. Even if you don't have time to describe what the extension does and how to use it, at least a few screenshots are needed so that people at least can guess what it can do.

adamrankin:
Can we integrate this into the index?

lassoan:
Merged. Thanks for your contribution!

I'm sorry that we had to ask for improvements, we just try to maintain some minimal standard to only let things in the Extension manager that are actually usable. This often requires some amount of documentation.

jcfr:
I added a README to the extension and also cleaned up the build system of SlicerVAAST and associated dependencies.

xref https://github.com/VASST/SlicerVASST/pull/4
xref https://github.com/openigtlink/SlicerOpenIGTLink/pull/11
xref https://github.com/openigtlink/OpenIGTLink/pull/174

PR #1529, opened 2018-03-01
WIP: add a script to update master to actual hash

fedorov:
Follow up on discussion here:

https://discourse.slicer.org/t/extensionindex-4-8-branch-is-using-master-for-many-most-extensions/2222

Tested and confirmed to work with the current content of the repo. Status after running the script:

image

Excerpt from `git diff`:

image

### DIFF
--- a/util/get_hash.py
+++ b/util/get_hash.py
@@ -1,0 +1,56 @@
+from urllib2 import urlopen, Request
+import json, os, glob
+
+token = "my personal access token - GitHub will throttle down requests without it"
+headers = {"Authorization": 'token %s' % token}
+
+def main():
+  # loop over all s4ext in the input directory
+  for s4extName in glob.glob("*.s4ext"):
+
+    print("Processing "+s4extName)
+
+    fileContent = []
+    org = None
+    repo = None
+    hash = None
+
+    with open(s4extName,'r') as s4extFile:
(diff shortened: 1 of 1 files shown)


jcfr:
If using python, I suggest to add the script directly into Slicer (in `Slicer/Base/Python/slicer/release`) and make use of [ExtensionDescription](https://github.com/Slicer/Slicer/blob/master/Utilities/Scripts/SlicerWizard/ExtensionDescription.py) which already implement read/write of description file.

Also, I suggest to keep master the way it is. Having `master` specified as `scmrevision` is convenient for teams who chose to do so.

Instead, the changes should be pushed to a branch named `master-X.Y` as explained in https://discourse.slicer.org/t/extensionindex-4-8-branch-is-using-master-for-many-most-extensions/2222/9?u=jcfr

Using the shell script, update of the file is quite simple too:

```
[...]
latest_scmrevision=$(git rev-list -n1 HEAD)
sed -e "s/scmrevision.*/scmrevision ${latest_scmrevision}/" -i ${extensionName}.s4ext
[...]
```

fedorov:
Thanks for the feedback. I replied on the discourse thread.

fedorov:
@jcfr any thoughts on this PR?

PR #1537, opened 2018-03-12
Update SlicerIGT.s4ext

adamrankin:
Updating git url to enable build

### DIFF
--- a/SlicerIGT.s4ext
+++ b/SlicerIGT.s4ext
@@ -6,7 +6,7 @@
 
 # This is source code manager (i.e. svn)
 scm git
-scmurl https://github.com/SlicerIGT/SlicerIGT.git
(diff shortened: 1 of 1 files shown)


lassoan:
Have you had problems with checking out a repository with https?

adamrankin:
Not with updated Git for Windows. But the build machines are failing to pull any extension defined with https.

lassoan:
77 extensions use `scmurl https://github.com...`. I don't think we should change all of them. What build machines don't have https?

jcfr:
While doing some cleanup, we removed private SSH keys that should not be on that machine.  I will locally generate a pair of key for that machine and it should allow anonymous https checkout to complete without asking credentials.

PR #1542, opened 2018-03-22
update PETTumorSegmentation

chribaue:


lassoan:
This pull request updated the git has of a wrong extension. I've fixed the issue: https://github.com/Slicer/ExtensionsIndex/commit/7464274512ef5b8dd5137394d1e2334e15d2f016

Manual updates of git hashes takes time and error-prone. You can make copy-paste errors, update the wrong s4ext file or the wrong branch of the ExtensionIndex, etc. It should be avoided.

If for some reason you feel safer with having specific git hashes in the s4ext file then the update should be fully automatic. You can use @jcfr's extension wizard scripts.

However, a much cleaner and better controllable workflow is to use protected release branches and set the release branch name in the s4ext file instead of a specific hash. The release branch can be master or any other name. Protected branches can be configured to require reviews and be writeable only by certain users. Therefore, instead of enforcing Slicer core developers to click blindly on merging a new git hash, you can control what gets into the release in your own repository by defining your own rules that you set for the protected branch.

ihnorton:
Another potential approach is a release bot. There are a few around, but the one I'm most familiar with is "attobot" for Julia packages. It is a travis hook that runs on AWS Lambda; essentially watches any release-tags on a repository, and automatically makes a validated PR to the [central index](https://github.com/JuliaLang/METADATA.jl). Automated update was imperative for Julia with [close to 2000 packages](https://pkg.julialang.org/pulse.html), because full manual review of every update PR by only 3-4 people was starting to be a big time-sink and blocker.  It's not so much an issue for ~100 extensions, but clean, validated automation is always nice in any case. The [code](https://github.com/attobot/attobot/blob/master/AttoBot/lambda_function.py) is quite short and MIT licensed, so it wouldn't be too crazy to just adapt it (if/when we have full continuous integration for extensions).

lassoan:
Thank you. Let's continue discussion here: https://discourse.slicer.org/t/slicer-extension-index-update-procedure/2415

chribaue:
Thank you for catching this! I'll switch our extensions to a 
> protected release branches and set the release branch name in the s4ext file

This will save you guys time as well

PR #1548, opened 2018-04-11
ENH: created new extension `SlicerLayoutButtons`

che85:


pieper:
LGTM, but adding a selftest would be nice.

che85:
Makes sense. Will do.

che85:
@pieper Selftests are added. Can anyone merge this?

PR #1549, opened 2018-05-07
Update OsteotomyPlanner.s4ext

sjh26:


OsteotomyPlanner.s4ext
```diff
@@ -7,15 +7,15 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl git://github.com/KitwareMedical/OsteotomyPlanner.git
-scmrevision 8258e52
+scmrevision ExtensionManager
```
fedorov: This is a bit unusual name for a branch - how about "Slicer-nightly" or something like this? Just a suggestion. If you prefer your name - that is fine with me.

OsteotomyPlanner.s4ext
```diff
@@ -7,15 +7,15 @@
...
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files
 # - The dependencies will be built first
 depends NA
 
 # Inner build directory (default is ".")
-build_subdirectory inner-build
+build_subdirectory .
```
fedorov: Can you confirm this is not a mistake? In some cases the use of "inner-build" implies superbuild organization of an extension, and changing this attribute can break the extension.

I won't make time to test this, so this is just a comment based on superficial review.

sjh26: Not a mistake!

PR #1552, opened 2018-05-30
BUG: Bump SlicerRT version to remove reference to SlicerMacroCheckExt…

gregsharp:
…ernalProjectDependency

### DIFF
--- a/SlicerRT.s4ext
+++ b/SlicerRT.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl https://github.com/SlicerRt/SlicerRT.git
-scmrevision 5a6dfa95fd323b8fd6109ad5a9f83edd5ed201be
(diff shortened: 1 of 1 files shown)


cpinter:
Apparently I merged the ExtensionIndex changes to the wrong branch the last time. To fix this:
1. Revert master ExtensionIndex SlicerRT.s4ext to use the SlicerRT master
2. Fix the 4.8 hash for the one I accidentally did for the nightly

SlicerRT used to use branches until the latest release. As we are expecting more frequent stable releases (apparently not quite happening), and SlicerRT development considerably slowed down, there is no real motivation to keep a branch. After Jc's CMake "modernization" changes, SlicerRT build broke, and that's why I wanted to just fix the hash for 4.8.

cpinter:
I changed the master ExtensionIndex to use the master SlicerRT, so the nightly will always use tha latest version, no need to update hashes.

I also created a master-48 SlicerRT branch which I reset to the revision that still works with 4.8. I made the 4.8 ExtensionIndex use that branch.

PR #1556, opened 2018-06-11
BUG: SlicerProstateAblation had incorrect dependency

sjh26:
This was crashing the nightly extension builds

### DIFF
--- a/SlicerProstateAblation.s4ext
+++ b/SlicerProstateAblation.s4ext
@@ -12,7 +12,7 @@ scmrevision master
 # list dependencies
 # - These should be names of other modules that have .s4ext files
 # - The dependencies will be built first
-depends SlicerDevelopmentToolbox SlicerZFrameRegistration
+depends SlicerDevelopmentToolbox ZFrameRegistration
 
 # Inner build directory (default is ".")
 build_subdirectory .


sjh26:
@cpinter

jcfr:
That is good as a short term fix.

I suggest @leochan2009 rename the project in the [CMakeLists.txt](https://github.com/SlicerProstate/SlicerZFrameRegistration/blob/28d43dd19c5052dbcc51ff79a108f91f32e5ea16/CMakeLists.txt#L3) and then rename the corresponding description file and usage in dependencies.

leochan2009:
@jcfr , Thanks for the suggestion, i think the name "SlicerZFrameRegistration" is more clear.
I will update the project name and all projects that dependent on SlicerZFrameRegistration

Best,
Longquan

fedorov:
@jcfr I am not sure renaming is a good idea. It seems like the convention in naming extensions is NOT to use "Slicer" as prefix - you can look at the content of ExtensionIndex. I am a bit worried that instead of this one simple fix that was merged in this PR we instead start changing many things in many places, and this can get confusing. Considering @leochan2009's last day here at BWH is this coming Thursday,  I would prefer not to change things that are working. What do you think?

jcfr:
> What do you think?

Here is the current documentation https://www.slicer.org/wiki/Documentation/Nightly/Developers/FAQ#Should_the_name_of_the_source_repository_match_the_name_of_the_extension_.3F

Either way would work.

>  last day here at BWH is this coming Thursday

In that case, let's simply rename the reference in the description file to `SlicerZFrameRegistration` as originally proposed by @sjh26 and keep the name of the CMake project as `ZFrameRegistration. `

lassoan:
I think the most common pattern is use Slicer(ExtensionName) as GitHub repository name, and using ExtensionName as extension name. We only keep the Slicer prefix in the extension name when the extension is a Slicer interface to some third-party library (SlicerOpenIGTLink, SlicerElastix, SlicerOpenCV, …).

jcfr:
> I think the most common pattern

I updated the FAQ entry to include your comment.

fedorov:
> I think the most common pattern is use Slicer(ExtensionName) as GitHub repository name, and using ExtensionName as extension name.

That was my impression too. Thanks!

PR #1564, opened 2018-07-28
Updating repository name and URL

adamrankin:


lassoan:
It would be better to find a more descriptive name. As is, it is unlikely that anyone would discover your extension if not explicitly searching for it.

PR #1566, opened 2018-08-10
Adding SlicerCamera extension

adamrankin:
For nightly release

### DIFF
--- a/SlicerCamera.s4ext
+++ b/SlicerCamera.s4ext
@@ -1,0 +1,43 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl https://github.com/VASST/SlicerCamera.git
+scmrevision 57ccab548314c308cd11b8282296e5eb0ae539f3
+
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends SlicerIGT SlicerOpenCV VASSTAlgorithms
+
+# Inner build directory (default is ".")
+build_subdirectory .
+
+# homepage
+homepage https://github.com/VASST/SlicerCamera
+
+# Firstname1 Lastname1 ([SubOrg1, ]Org1), Firstname2 Lastname2 ([SubOrg2, ]Org2)
+contributors Adam Rankin (Robarts Research Institute), Elvis Chen (Robarts Research Institute)
+
+# Match category in the xml description of the module (where it shows up in Modules menu)
+category Webcams
+
+# url to icon (png, size 128x128 pixels)
+iconurl https://github.com/VASST/SlicerCamera/blob/master/SlicerCamera.png
+
+# Give people an idea what to expect from this code
+#  - Is it just a test or something you stand behind?
+status Stable
+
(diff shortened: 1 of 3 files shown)


jcfr:
Thanks for the contribution.

Few comments .. 

Should the extension be renamed `SlicerVideoCamera` ? 
Or should the logo be updated ? 

* [ ] update screenshot (currently link to http://www.example.com/Slicer/Extensions/SlicerCamera/Screenshots/1.png)

As a side note, would it be possible to link a 128x128 version of the logo.

lassoan:
It looks all good except the extension name should be more specific. For example VideoCamera or CameraCalibration would be good extension names. Recommended repository name is the extension name prefixed with "Slicer".

adamrankin:
Would SlicerCameras be ok (extension name "Cameras")? I know not much difference between Webcams and Cameras but in these tools apply to still cameras as well as video cameras.

lassoan:
Camera term is already taken by VTK and Slicer core. For me, VideoCamera term is distinct enough and indicates that it is about optical cameras and not about virtual rendering cameras. Maybe we can find other terms, if you don't like VideoCamera.

adamrankin:
Agreed, sticking with `Webcam`

lassoan:
Sorry for taking so much time with finding a good name, but a good name makes a huge difference, and it is hard to change in the future (it cause errors in dependent extensions, users cannot find it, extension statistics cannot be aggregated, etc).

Are the algorithms specify developed to work on webcams (i.e., low-cost, low-quality video cameras) or they are applicable to high-end optics as well (endoscope, DSLR, etc)?

pieper:
CameraModels or CameraModeling would give me a better idea of what it is.

adamrankin:
Well, the key question is, is SlicerCameras out? Because that is the best name for the extension, as it is focused around calibration, use, and manipulation of cameras. 

Second best would be Webcams, but that incorrectly dissuades from DSLRs, endoscopes, etc... which are valid candidates for the tools within.

fedorov:
Looking from the outside, after reading this discussion, I find the SlicerWebcams name rather misleading.

fedorov:
+1 for CameraCalibration

pieper:
I like CameraCalibration.  Another option might be CameraMathematics

adamrankin:
The extension is more than just calibration though. VideoCameras is also misleading as it works for still cameras as well.

PinholeCameras is not bad.

What about Optics (SlicerOptics repo name)? Or Optical Cameras?

adamrankin:
Ok, final decision of VideoCameras (repo SlicerVideoCameras)?

adamrankin:
@jcfr I think you're the final say on naming.

adamrankin:
Ok, renamed. If possible could this be merged before tomorrow's nightly build?

lassoan:
Icon and screenshot URLs are invalid (for example, correct icon URL would be: https://raw.githubusercontent.com/VASST/SlicerVideoCameras/master/SlicerVideoCameras_128.png). Could you please fix them?

lassoan:
Merged. Thank you for your contribution!

Note that the git hash was hard-coded in the s4ext file, so your latest fixes are not included. I would recommend to replace the git hash by branch name (master) to avoid the need of manually updating the extension index each time you change your extension.

PR #1572, opened 2018-09-04
Add SlicerWMA python (installer only) extension

ihnorton:
This is not a "functional" extension yet: all it does is package [WhiteMatterAnalysis](https://github.com/SlicerDMRI/whitematteranalysis) and dependencies. We want to add it now in order to:
- use the package with full dependencies already installed, within Slicer's python. This is for a potentially low-connectivity environment (want to be able to hand out self-contained extension .tar.gz locally if needed)

- do further development/testing of (scripted) wrapper tooling on multiple platforms without setting up a full build environment

The extension logo clearly marks as WIP:

logo

### DIFF
--- a/SlicerWMA.ext
+++ b/SlicerWMA.ext
@@ -1,0 +1,25 @@
+scm git
+scmurl https://github.com/SlicerDMRI/whitematteranalysis
+scmrevision master
+
+depends NA
+
+build_subdirectory inner-build
+
+homepage http://slicerdmri.github.io
+
+contributors Lauren O'Donnell, Fan Zhang (Brigham & Women's Hospital), and others.
+
+category Diffusion
+
+iconurl https://gist.githubusercontent.com/ihnorton/f860e58c99015a6b1a3b04cc4c772412/raw/b2a440420cc01630b339c52b3feaa2744eae7039/SlicerWMA.png
+
+status WIP
+
+# One line stating what the module does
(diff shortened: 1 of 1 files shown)


ihnorton:
cc @ljod @zhangfanmark

ljod:
I agree that it will be nice to put a more reasonable icon that looks more positive. We are making progress!

ihnorton:
> There are other "utility" extensions in the app store already (e.g., SlicerOpenCV)

Thanks, that is a good analogy -- essentially what we want here is a library package right now so that we can use the command line scripts with Slicer's python.

> the icon needs to be fixed

Done (usual "puzzle" icon for now because I want to test this in the nightly, but I will make a nicer one)

> maybe a description of the state of the extension and future plans 

Added to: https://github.com/SlicerDMRI/whitematteranalysis/blob/master/README.md

> By the way, why do you need to install another Python? How it is used?

There is no extra python -- I guess the "(installer only)" in the title was misleading. All the extension does is install the upstream whitematteranalysis code and (several) dependencies.

jcfr:
Nitpick: If this could be a pip install-able package, the extension itself would be a simple external project.

ljod:
That sounds like a good idea. I think we should do that for the future. Thanks!

ihnorton:
@jcfr did this miss a cutoff time for the build system? It doesn't seem to have tried to build this last night (at least there is no log I can find).

jcfr:
The change made it, it wasn't picked up because of typo in the file name.

`SlicerWMA.ext` -> `SlicerWMA.s4ext`

ihnorton:
Thanks

PR #1574, opened 2018-09-05
ci: Add continuous integration checking for unexpected files

jcfr:


jcfr:
@ihnorton This will prevent future regression like the one you observed. It will detect if there invalid names and list them,

jcfr:
It is currently failing and will be expected to pass after I rebase the topic.

jcfr:
Error detection work as expected:

```
Looking for unexpected files
./SlicerWMA.ext
Looking for unexpected files - failed
Exited with code 1
```

PR #1578, opened 2018-09-05
ENH: Add s rep extension

ZhiyLiu:
The full name of this extension is skeletal representation extension, focusing on shape analysis based on s-rep model.


ZhiyLiu:
Move to #1582

PR #1586, opened 2018-09-25
Update SlicerCochlea.s4ext 4.8

idhamari:


idhamari:
Sorry, for some reason, I just see your reply now. I tested it using Slicer 4.8.1 and it works.

jcfr:
Thanks for following up.

Since we released Slicer 4.10 (announcement will be out soon), check if you need to update  `SlicerCochlea.s4ext` in the `4.10` branch of the extension index. 

In the mean team, but no new extension for Slicer 4.8 will be built.

idhamari:
Thanks for your effort and support. I just saw that. I will do more updating and testing soon using 4.10 and 4.11.

PR #1590, opened 2018-10-09
Add SlicerDiffusionQC extension

ihnorton:
"SlicerDiffusionQC provides a quality-estimation algorithm for diffusion-weighted MRI gradient images, paired with an interactive, graphical review and approval tool."


pieper:
LGTM

lassoan:
If the checklist is complete then the extension is ready for integration:

- [ ] Extension has a reasonable name (not too general, not too narrow, suggests what the extension is for)
- [ ] Repository name is Slicer+ExtensionName
- [ ] Extension description summarizes in 1-2 sentences what the extension is usable (should be understandable for non-experts)
- [ ] Extension URL and revision (scmurl, scmrevision) is correct, consider using a branch name (master, release, …) instead of a specific git has to avoid re-submitting pull request whenever the extension is updated
- [ ] Extension icon URL is correct
- [ ] Screenshot URLs (screenshoturls) are correct, contains at least one
- [ ] Homepage URL points to valid webpage containing the followings:
  - [ ] Extension name
  - [ ] Short description: 1-2 sentences, which summarizes what the extension is usable for
  - [ ] At least one nice, informative image, that illustrates what the extension can do. It may be a screenshot.
  - [ ] Description of contained modules: at one sentence for each module
  - [ ] Tutorial: step-by-step description of at least the most typical use case, include a few screenshots, provide download links to sample input data set

lassoan:
Notes:
- Repository name should start with Slicer (and it is, so it is good). However, extension name should only start with Slicer, if otherwise it would be confusing (typically when we refer to another library, such as Elastix or Matlab; or a large field, such as RT, Heart, IGT). Overall, in this case, I think it would be better to name the extension as DiffusionQC.
- Most EXTENSION_ fields in https://github.com/pnlbwh/SlicerDiffusionQC/blob/master/CMakeLists.txt are invalid (SlicerDMRI is not listed as a dependency, screenshot url, homapage are invalid, etc.)
- Extension icon is invalid (it's placeholder icon) - https://raw.githubusercontent.com/pnlbwh/SlicerDiffusionQC/master/SlicerDiffusionQC.png
- Screenshot URLs are missing
- Images are missing from the extension documentation page (it is extremely important if you want people to try your extension)
- Tutorial is nice but no links are provided to sample data sets. If users run into issues they will not know if the software works at all or only their data is not suitable and most likely will simply give up. 
- You may consider moving "File description" section under a new "Developer" section, as most users are not interested in what is in the extension but how to use it.

ihnorton:
Thanks. All changes addresses except tutorial, Tashrif will look at that when he gets back.

lassoan:
Thanks for the updates. I see that you've updated the icon URL, but screenshot URLs line is still missing.

PR #1596, opened 2018-12-04
add lapdMouseBrowser

chribaue:


fedorov:
Looks cool, thanks!

Looking at the circleci output for the failing test, it appears that "Slicer" prefix is now required for the extensions, so you would either need to change the repo name, or make an argument why you would prefer to keep the existing repo name.

fedorov:
Looks good to me! :+1:

lassoan:
Thank you very much for your contribution, it looks really nice! The extension's documentation is exemplary, too.

There is just one minor thing that could make it a bit easier for people to understand what your extension is about: please dd some more detail to the extension description. Currently it is just "Tumor and lymph node segmentation in PET scans". Instead, you could use the same intro sentence as on the extension's website: "The Lung Anatomy + Particle Deposition Mouse Archive (lapdMouse) contains anatomically derived lung models and aerosol deposition measurements of mice for modeling and computational toxicology.")

chribaue:
Andras, good catch. That was a copy/paste error. I issued a PR for an update: https://github.com/Slicer/ExtensionsIndex/pull/1598

PR #1597, opened 2018-12-04
ENH: Automatic tractography extension with deep learning

juanprietob:
First commit for extension AutoTract

### DIFF
--- a/RegQAExtension.s4ext
+++ b/RegQAExtension.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl https://github.com/gsi-biomotion/Slicer-RegQA.git
-scmrevision ada7f66
+scmrevision master
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files


juanprietob:
Do I have to rename the repository?

fedorov:
> This app functions also as a standalone package, are you guys requiring that the extension is named Slicer+name?

This is a good point. I personally do not think Slicer prefix should be mandatory. Also, there are many many existing extensions that do not have Slicer prefix. I don't quite see why Slicer would be in position to mandate naming of the repositories maintained by the extension developers. In some cases it is natural and very appropriate, in other cases, like this one, it is opposite from helpful.

jcfr:
> This app functions also as a standalone package

This is a valid exception, I suggest we update the checklist.

@juanprietob Could you add the [3d-slicer-extension](https://github.com/topics/3d-slicer-extension) GitHub topic to the repository ? 

>  I don't quite see why Slicer would be in position to mandate naming of the repositories maintained by the extension developers

Generally speaking, it makes sense to suggest to have the "Slicer" prefix.

I also think promoting the Slicer brand is a reasonable and strategic move to help secure follow up funding.

fedorov:
@jcfr yes, I totally agree it definitely makes sense to suggest, inform and encourage the developers to adopt the prefix, but I think in the end the developer should make the decision. At the moment, it is hard-coded in the test, and the failing test makes it look as if this is a mandatory requirement.

image

If this is not a mandatory requirement, but a suggestion, the failing test message should probably be revised to something like "Please consider adopting the Slicer prefix, or provide your argument for not using it in the PR discussion".

juanprietob:
> > This app functions also as a standalone package
> 
> This is a valid exception, I suggest we update the checklist.
> 
> @juanprietob Could you add the [3d-slicer-extension](https://github.com/topics/3d-slicer-extension) GitHub topic to the repository ?

Done!

> 
> > I don't quite see why Slicer would be in position to mandate naming of the repositories maintained by the extension developers
> 
> Generally speaking, it makes sense to suggest to have the "Slicer" prefix.
> 
> I also think promoting the Slicer brand is a reasonable and strategic move to help secure follow up funding.

jcfr:
Here is a followup topic related to improvement of the contribution checklist. See https://discourse.slicer.org/t/extension-contribution-checklist/4975

jcfr:
Could you rebase your change instead of merging master into your branch, it makes it quite difficult to understanding what is happening. Thanks again.

juanprietob:
I made a mistake when updating my repository and lost the commits done on this threat. 
This is the new pull request https://github.com/Slicer/ExtensionsIndex/pull/1607

PR #1610, opened 2019-01-28
Add SlicerDcm2nii extension

ihnorton:
- [ ] Screenshot URLs (screenshoturls) are correct, contains at least one
- [ ] Homepage URL points to valid webpage containing the following:
  - [ ] Extension name
  - [ ] Short description: 1-2 sentences, which summarizes what the extension is usable for
  - [ ] At least one nice, informative image, that illustrates what the extension can do. It may be a screenshot.
  - [ ] Description of contained modules: at one sentence for each module
  - [ ] Tutorial: step-by-step description of at least the most typical use case, include a few screenshots, provide download links to sample input data set

### DIFF
--- a/SlicerDcm2nii.s4ext
+++ b/SlicerDcm2nii.s4ext
@@ -1,0 +1,23 @@
+scm git
+scmurl https://github.com/SlicerDMRI/SlicerDcm2nii.git
+scmrevision master
+
+depends NA
+
+build_subdirectory inner-build
+
+homepage http://slicerdmri.github.io
+
+contributors Isaiah Norton, Lauren O'Donnell (Brigham & Women's Hospital), and others.
+
+category Utilities
+
+iconurl iconurl https://raw.githubusercontent.com/SlicerDMRI/SlicerWMA/master/SlicerWMA.png
+
+status Beta
+
+# One line stating what the module does
+description SlicerDcm2nii builds and distributes dcm2niix (https://github.com/rordenlab/dcm2niix) as part of the Slicer superbuild.
+
+# 0 or 1: Define if the extension should be enabled after its installation.
+enabled 1


ihnorton:
This extension build dcm2niix binary as part of the Slicer superbuild. See [this discussion](https://discourse.slicer.org/t/convert-single-files-into-one-nii/5407/12?u=ihnorton) for background.

> A quick but useful initial step could be to provide dcm2niix based DICOM plugin in an extension.

fedorov:
Thanks Isaiah! I personally would add it as a DICOM plugin, but I am not the one doing the work, so have no right to complain!

pieper:
Yes, this looks like a great step in the right direction. 👍

lassoan:
Thank you Isaiah, nice work!

> I personally would add it as a DICOM plugin, but I am not the one doing the work, so have no right to complain!

Fully agree. Ultimately, this will have to be exposed as a DICOM plugin, but the DICOM plugin can be implemented as a Python script that uses a CLI. This is how CreateDICOMSeries CLI is used in the scalar volume plugin for data export.

ihnorton:
> I personally would add it as a DICOM plugin, but I am not the one doing the work, so have no right to complain!

I agree, I just don't have time to do it right now and need to prioritize making it available for DWI-specific use for a collaborator.

We would be happy to have the functionality in this repo, or transfer the repo to another organization if that makes sense. For the time being, I added @fedorov @lassoan @hjmjohnson as admins on this repo.

ihnorton:
Seeing no objections or failures, I'm going to go ahead and merge this to get CI tonight.

fedorov:
@ihnorton 💯 👍 

I greatly appreciate the effort! Thank you for this contribution.

PR #1634, opened 2019-05-23
PULL_REQUEST_TEMPLATE: Fix link and improve wording

jcfr:


jcfr:
I see ... in that case, may be we could completely remove the bracket then ? 

> shown to the user as plain text in the editor

When you say "text editor", do you mean the one that is shown when creating a PR using https://hub.github.com/ ? 

May be we could have a redirect from `https://extensions.slicer.org/contribute` to https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/BuildTestPackageDistributeExtensions

lassoan:
Yes, I meant that when you create a pull request then the template is shown as plain text there, so the brackets are not useful.

For describing contribution guidelines, I think we should use a standard github.com/slicer/slicer(or ExtensionsIndex)/contribute.md. Using MediaWiki would be quite unusual/old-school for this nowadays.

PR #1655, opened 2019-08-03
add SegmentEditorSRSFilter.s4ext

sebastianandress:
  - [x] Extension name
  - [ ] Publication: link to publication and/or to PubMed reference (if available)
  - [x] License: We suggest you use a permissive license that includes patent and contribution clauses.  This will help protect developers and ensure the code remains freely available.  We suggest you use the [Slicer License](https://github.com/Slicer/Slicer/blob/master/License.txt) or the [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0). Always mention in your README file the license you have chosen.  If you choose a different license, explain why to the extension maintainers. Depending on the license we may not be able to host your work.
Read [here](https://opensource.guide/legal/#which-open-source-license-is-appropriate-for-my-project) to learn more about licenses.


sebastianandress:
I didn't, but now it is tested and ready for Python3.

We were quite unsure with the naming. Should we call it SRSS (Shrinkwrap-Raycast-Shrinkwrap-Solidify) filter? Any better suggestions?

pieper:
Nice contribution!

+1 for a more suggestive name - maybe something like ShrinkCastSegmentation (leaves out some details but might give people the general idea)

sebastianandress:
I really like ShrinkCast, should we call it ShrinkCast Surface Filter? But again it leaves out the solidifying. But I'd be fine with it 👍

lassoan:
The name should suggest *what* the effect does, so that users have a chance to find your effect when they need it. So, the name should contain something about filling holes, making solid, etc.

sebastianandress:
Maybe we should go for Shrink-Solidifying, since the Raycasting-Step also kind of "shrinkes" the sphere to the Model. Again, a simplification, but maybe the most understandable. What is your opinion?

pieper:
Maybe SolidifySurface?

lassoan:
> Maybe SolidifySurface?

It may be somewhat misleading, since it actually solidifies a segment. What do you think about `SlicerSolidifySegment` for extension name and `Solidify` for effect name?

sebastianandress:
Sorry for the late reply. I'm still not sure about the name. It also creates a model out of the segment if selected, in fact, for our patient study we used only this option.

Would you be good with Surface Solidify?

lassoan:
> It also creates a model out of the segment if selected, in fact, for our patient study we used only this option.

Are you aware of the "Export to files" feature in Segment Editor/Segmentations module, which allows writing STL or OBJ file directly from a segmentation? Are you sure you need an additional step of creating a model node if the goal is to have a surface mesh file?

If you don't want to export to file but you need the model nodes in the scene then you can export all the segments to models using two clicks in Data module or in Segmentations module. Are you sure you would want to add another export option to your effect? It would just make the user interface of your effect more complicated and inconsistent with all other effects. I understand if you need to minimize number of clicks for your users, but then you can add a new module that has an embedded Segment Editor widget and fully customized for your workflow. See for example QuickSegment slicelet [here](https://www.slicer.org/wiki/Documentation/Nightly/Developers/Slicelets).

sebastianandress:
Actually we added this option since the segmentation can only be so thick depending on the spacing of the related volume. And for 3D printing in our case sometimes thinner solidifications were needed.

But I'm fine with just calling it Solidify as you suggested above. I'd change it the following days.

lassoan:
I see, thank you, that's a good point, then your segment editor effect is actually different than the rest of the effects, so it makes sense that it is not consistent how others work. We should keep direct model export option (or export to another segmentation node, with closed surface representation) available.

Names could be:
- extension: SlicerSurfaceSolidify
- module: SegmentEditorSolidify
- effect: Solidify

lassoan:
I hope we did not discourage you too much with all our requests for changes. Users would benefit a lot from your extension, so it would be great if you could finalize the names and then we would add the extension to the index. Thanks again for your contribution!

sebastianandress:
Sorry for the late reply, I've been in holiday the past days. I like the suggestions, but I'd add Wrap in front of solidify, since it makes the technique more clear with its drawback: It won't solidify inner structures (to a specific point depending on the user preferences).
So the names would be:
- github repository: SlicerSurfaceWrapSolidify
- extension name: SurfaceWrapSolidify
- module: SegmentEditorWrapSolidify
- effect: Wrap solidify

lassoan:
Sounds good! Thanks for your contribution and patience with all these details. Please let us know when the name updates are ready!

lassoan:
Just two very small things:
- Category name should be "Segmentation" (not "Segmentations" - that would create an entire new category in the extension index)
- We put a small segment editor "badge" in the extension's icon if it contains segment editor effect (see [example](https://github.com/lassoan/SlicerSegmentEditorExtraEffects/blob/master/SegmentEditorExtraEffects.png) and [GIMP file for the badge](https://github.com/lassoan/SlicerSegmentEditorExtraEffects/blob/master/SegmentEditorExtraEffects.xcf)). Users could find out that your extension provides segment editor effect if you added this badge to your extension, too.

sebastianandress:
It should be fine now, the names are updated.

lassoan:
Thanks a lot, we are really close...
- Extension name is now "Slicer-SurfaceWrapSolidify" should be "SlicerSurfaceWrapSolidify".
- Module folder name is now "SurfaceWrapSolidify" should be "SegmentEditorWrapSolidify" (need to update in CMakeLists.txt, too)

sebastianandress:
Once again I added the changes, sorry for the confusion!

lassoan:
Merged! Thank you for your contribution and for your patience with all these small adjustments. Check the [dashboard](http://slicer.cdash.org/index.php?project=SlicerPreview) tomorrow if build was successful on all platforms.

jcfr:
Ditto. Thanks @sebastianandress :pray:

sebastianandress:
Very nice, also thank you for your patience!

sebastianandress:
Which modules should I remove from the dependencies list? All of these are built in, isn‘t it:
SegmentEditor Segmentations Models SegmentStatistics

Of course I also will close the brackets, sorry for that.

lassoan:
You need to list extensions you depend on and not modules. Since you don't depend on any extension, you need to put just NA there:

    set(EXTENSION_DEPENDS "NA")

sebastianandress:
I changed it in CMakeList.txt files as well as in the Python scripts and pushed the repository. Should I also create a new pull request for the s4ext file or is that ok?

lassoan:
@sjh26 had already fixed the s4ext file in this repository, so we should be good. We'll see the dashboard tomorrow.

PR #1660, opened 2019-10-23
ENH:  Add check to verify that the listed scm is not "local"

sjh26:
Adds a check to make sure that the scm tag is not "local"  which won't work on the factory systems.

See: https://github.com/Slicer/ExtensionsIndex/commit/ad2b02223a4e7a3a2070f821cad52daacc91390c

Also, previously the check_scmurl_syntax check was only done when we weren't doing the repo name check, which doesn't seem right.  With this commit, the repo check is done based on the parameters, and then the scmurl check is done every time.

### DIFF
--- a/scripts/check_description_files.py
+++ b/scripts/check_description_files.py
@@ -70,6 +70,11 @@ def check_scmurl_syntax(extension_name, metadata):
             extension_name, check_name,
             "scmurl scheme is '%s' but it should by any of %s" % (scheme, supported_schemes))
 
+@require_metadata_key("scm")
+def check_scm_notlocal(extension_name, metadata):
+    check_name = "check_scm_notlocal"
+    if metadata["scm"] == "local":
+        raise ExtensionCheckError(extension_name, check_name, "scm cannot be local")
 
 @require_metadata_key("scmurl")
 @require_metadata_key("scm")
@@ -146,6 +151,7 @@ def main():
     if not checks:
         checks = [
             check_scmurl_syntax,
+            check_scm_notlocal,
         ]
 
     total_failure_count = 0


scripts/check_description_files.py
```diff
@@ -110,10 +115,10 @@ def main():
     if args.check_git_repository_name:
         checks.append(check_git_repository_name)
 
-    if not checks:
```
jcfr: If `checks` was false, it means no argument was passed and the default list of check was applied. This was done on purpose.

What about adding a parameter `--check-scm-not-local`

sjh26: Okay, I understand better what is happening now.  You are using the same script for two different checks, so the default is run every time for one of the checks.  Do we want to have this as a separate check or just add it to the default checks?  If we add it to the default checks we won't need a parameter.

jcfr: Systematically adding an argument to run a specific test makes it easy to test the effect of a specific check.

To simplify the maintenance of the script, I am also open to the idea of removing the argument parser and systematically calling all test functions.

lassoan:
@sjh26 What is the status of this? Is this still needed/relevant?

jcfr:
After applying the following patch locally and running the script, it effectively detects the use of `local`


### Test

```bash
$ python3 ./scripts/check_description_files.py \
  --check-dependencies . \
  $(find . -name "*.s4ext")
SequenceRegistration.s4ext
  scm cannot be local
SlicerFab.s4ext
  scm cannot be local
Checked content of 223 description files.
Checked dependency between 178 extensions.
Total errors found in extension descriptions: 2
```


### Patch

```diff
diff --git a/SequenceRegistration.s4ext b/SequenceRegistration.s4ext
index 970ac6e..bffb17a 100644
--- a/SequenceRegistration.s4ext
+++ b/SequenceRegistration.s4ext
@@ -5,7 +5,7 @@
 #
 
 # This is source code manager
-scm git
+scm local
 scmurl https://github.com/moselhy/SlicerSequenceRegistration
 scmrevision master
 
diff --git a/SlicerFab.s4ext b/SlicerFab.s4ext
index df27719..e75b137 100644
--- a/SlicerFab.s4ext
+++ b/SlicerFab.s4ext
@@ -5,7 +5,7 @@
 #
 
 # This is source code manager
-scm git
+scm local
 scmurl https://github.com/SlicerFab/SlicerFab
 scmrevision master

```

PR #1665, opened 2019-11-24
Add NVIDIA Clara AI-Assisted Annotation Extension to Slicer-4.11

lassoan:
This adds Nvidia's AI-assisted annotation (powered by Clara Train SDK) automatic deep learning based segmenter to Segment editor.

It will be functional once this pull request is merged, too:
https://github.com/NVIDIA/ai-assisted-annotation-client/pull/43


pieper:
@lassoan shall we set up an instance of clara running on a cloud server so we can demo this at rsna?

lassoan:
Currently, I use the server set up by Brett Marinelli, but it is not publicly available, so it would be great to set it up on a cloud server that could be used publicly.

PR #1671, opened 2019-12-31
Cdash errors for Slicer 4.10 extensions

jamesobutler:
This fixes various configure errors as seen on the Slicer4 extension cdash.


lassoan:
Thanks a lot for taking care of these.

PR #1689, opened 2020-03-19
add OpenDose extension

BishopWolf:
- [x] Repository is associated with `3d-slicer-extension` GitHub topic so that it is listed [here](https://github.com/topics/3d-slicer-extension). To edit topics, click "Manage topics" on your repository's main page, on the left, near the top (just below the repository description). To learn more about topics, read https://help.github.com/en/articles/about-topics
  - [x] Extension name
  - [ ] Publication: link to publication and/or to PubMed reference (if available)
  - [x] License: We suggest you use a permissive license that includes patent and contribution clauses.  This will help protect developers and ensure the code remains freely available.  We suggest you use the [Slicer License](https://github.com/Slicer/Slicer/blob/master/License.txt) or the [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0). Always mention in your README file the license you have chosen.  If you choose a different license, explain why to the extension maintainers. Depending on the license we may not be able to host your work.
Read [here](https://opensource.guide/legal/#which-open-source-license-is-appropriate-for-my-project) to learn more about licenses.

### DIFF
--- a/OpenDose4D.s4ext
+++ b/OpenDose4D.s4ext
@@ -7,7 +7,7 @@
 # This is source code manager (i.e. svn)
 scm git
 scmurl git://gitlab.com/opendose/OpenDose4D.git
-scmrevision develop
+scmrevision master-stable
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files
@@ -18,14 +18,14 @@ depends SlicerElastix
 build_subdirectory .
 
 # homepage
-homepage http://slicer.org/slicerWiki/index.php/Documentation/Nightly/Extensions/OpenDose4D
+homepage https://gitlab.com/opendose/opendose4d
 
 # Firstname1 Lastname1 ([SubOrg1, ]Org1), Firstname2 Lastname2 ([SubOrg2, ]Org2)
 # For example: Jane Roe (Superware), John Doe (Lab1, Nowhere), Joe Bloggs (Noware)
 contributors Alex Vergara Gil (INSERM, France), Janick Rueegger (KSA, Switzerland)
 
 # Match category in the xml description of the module (where it shows up in Modules menu)
-category Nuclear Medicine
+category Radiotherapy
 
 # url to icon (png, size 128x128 pixels)
 iconurl https://gitlab.com/opendose/opendose4d/-/blob/develop/Dosimetry4D/Resources/Icons/Dosimetry4D.png
--- a/OpenDose4D.s4ext
+++ b/OpenDose4D.s4ext
@@ -38,7 +38,7 @@ status stable
 description This module implements full 3D dosimetry for molecular radiotherapy on multiple time points.
 
 # Space separated list of urls
-screenshoturls
+screenshoturls https://gitlab.com/opendose/opendose4d/-/blob/develop/Dosimetry4D/Resources/Icons/ss1.png https://gitlab.com/opendose/opendose4d/-/blob/develop/Dosimetry4D/Resources/Icons/ss2.png
 
 # 0 or 1: Define if the extension should be enabled after its installation.
 enabled 1
--- a/OpenDose4D.s4ext
+++ b/OpenDose4D.s4ext
@@ -1,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl git://gitlab.com/opendose/OpenDose4D.git
+scmrevision develop
+
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends SlicerElastix
+
+# Inner build directory (default is ".")
+build_subdirectory .
+
+# homepage
+homepage http://slicer.org/slicerWiki/index.php/Documentation/Nightly/Extensions/OpenDose4D
+
+# Firstname1 Lastname1 ([SubOrg1, ]Org1), Firstname2 Lastname2 ([SubOrg2, ]Org2)
+# For example: Jane Roe (Superware), John Doe (Lab1, Nowhere), Joe Bloggs (Noware)
+contributors Alex Vergara Gil (INSERM, France), Janick Rueegger (KSA, Switzerland)
+
+# Match category in the xml description of the module (where it shows up in Modules menu)
+category Nuclear Medicine
+
+# url to icon (png, size 128x128 pixels)
+iconurl https://gitlab.com/opendose/opendose4d/-/blob/develop/Dosimetry4D/Resources/Icons/Dosimetry4D.png
+
+# Give people an idea what to expect from this code
+#  - Is it just a test or something you stand behind?
+status stable
+
+# One line stating what the module does
+description This module implements the full 3D Dosimetry for molecular radiotherapy on multiple time points.
+
+# Space separated list of urls
+screenshoturls
+
+# 0 or 1: Define if the extension should be enabled after its installation.
+enabled 1
--- a/OpenDose4D.s4ext
+++ b/OpenDose4D.s4ext
@@ -35,7 +35,7 @@ iconurl https://gitlab.com/opendose/opendose4d/-/blob/develop/Dosimetry4D/Resour
 status stable
 
 # One line stating what the module does
-description This module implements the full 3D Dosimetry for molecular radiotherapy on multiple time points.
+description This module implements full 3D dosimetry for molecular radiotherapy on multiple time points.
 
 # Space separated list of urls
 screenshoturls
--- a/OpenDose4D.s4ext
+++ b/OpenDose4D.s4ext
@@ -28,7 +28,7 @@ contributors Alex Vergara Gil (INSERM, France), Janick Rueegger (KSA, Switzerlan
 category Radiotherapy
 
 # url to icon (png, size 128x128 pixels)
-iconurl https://gitlab.com/opendose/opendose4d/-/blob/develop/Dosimetry4D/Resources/Icons/Dosimetry4D.png
+iconurl https://opendose.org/themes/default/images/opendose_favicon.png
 
 # Give people an idea what to expect from this code
 #  - Is it just a test or something you stand behind?
--- a/OpenDose3D.s4ext
+++ b/OpenDose3D.s4ext
@@ -6,7 +6,7 @@
 
 # This is source code manager (i.e. svn)
 scm git
-scmurl git://gitlab.com/opendose/OpenDose4D.git
+scmurl git://gitlab.com/opendose/OpenDose3D.git
 scmrevision master-stable
 
 # list dependencies
@@ -18,7 +18,7 @@ depends SlicerElastix
 build_subdirectory .
 
 # homepage
-homepage https://gitlab.com/opendose/opendose4d
+homepage https://gitlab.com/opendose/opendose3d
 
 # Firstname1 Lastname1 ([SubOrg1, ]Org1), Firstname2 Lastname2 ([SubOrg2, ]Org2)
 # For example: Jane Roe (Superware), John Doe (Lab1, Nowhere), Joe Bloggs (Noware)
@@ -38,7 +38,7 @@ status stable
 description This module implements full 3D dosimetry for molecular radiotherapy on multiple time points.
 
 # Space separated list of urls
-screenshoturls https://gitlab.com/opendose/opendose4d/-/blob/develop/Dosimetry4D/Resources/Icons/ss1.png https://gitlab.com/opendose/opendose4d/-/blob/develop/Dosimetry4D/Resources/Icons/ss2.png
+screenshoturls https://gitlab.com/opendose/opendose3d/-/blob/develop/Dosimetry4D/Resources/Icons/ss1.png https://gitlab.com/opendose/opendose3d/-/blob/develop/Dosimetry4D/Resources/Icons/ss2.png
 
 # 0 or 1: Define if the extension should be enabled after its installation.
 enabled 1


OpenDose4D.s4ext
```diff
@@ -0,0 +1,44 @@
...
+
+# url to icon (png, size 128x128 pixels)
+iconurl https://gitlab.com/opendose/opendose4d/-/blob/develop/Dosimetry4D/Resources/Icons/Dosimetry4D.png
+
+# Give people an idea what to expect from this code
+#  - Is it just a test or something you stand behind?
+status stable
+
+# One line stating what the module does
+description This module implements the full 3D Dosimetry for molecular radiotherapy on multiple time points.
```
lassoan:
```suggestion
description This module implements full 3D dosimetry for molecular radiotherapy on multiple time points.
```

OpenDose4D.s4ext
```diff
@@ -0,0 +1,44 @@
...
+
+# Firstname1 Lastname1 ([SubOrg1, ]Org1), Firstname2 Lastname2 ([SubOrg2, ]Org2)
+# For example: Jane Roe (Superware), John Doe (Lab1, Nowhere), Joe Bloggs (Noware)
+contributors Alex Vergara Gil (INSERM, France), Janick Rueegger (KSA, Switzerland)
+
+# Match category in the xml description of the module (where it shows up in Modules menu)
+category Nuclear Medicine
+
+# url to icon (png, size 128x128 pixels)
+iconurl https://gitlab.com/opendose/opendose4d/-/blob/develop/Dosimetry4D/Resources/Icons/Dosimetry4D.png
```
lassoan: You need to put image download URL here.

OpenDose4D.s4ext
```diff
@@ -0,0 +1,44 @@
...
+
+# homepage
+homepage http://slicer.org/slicerWiki/index.php/Documentation/Nightly/Extensions/OpenDose4D
+
+# Firstname1 Lastname1 ([SubOrg1, ]Org1), Firstname2 Lastname2 ([SubOrg2, ]Org2)
+# For example: Jane Roe (Superware), John Doe (Lab1, Nowhere), Joe Bloggs (Noware)
+contributors Alex Vergara Gil (INSERM, France), Janick Rueegger (KSA, Switzerland)
+
+# Match category in the xml description of the module (where it shows up in Modules menu)
+category Nuclear Medicine
```
lassoan: Use existing Radiotherapy category. If we have 5+ NM related extensions then we can create a new category for that.

OpenDose4D.s4ext
```diff
@@ -0,0 +1,44 @@
...
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends SlicerElastix
+
+# Inner build directory (default is ".")
+build_subdirectory .
+
+# homepage
+homepage http://slicer.org/slicerWiki/index.php/Documentation/Nightly/Extensions/OpenDose4D
```
lassoan: This should be your project homepage - https://gitlab.com/opendose/opendose4d

OpenDose4D.s4ext
```diff
@@ -0,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl git://gitlab.com/opendose/OpenDose4D.git
+scmrevision develop
```
lassoan: Probably you want to use master or release. You can merge commits into master or release branch when changes are ready.

OpenDose4D.s4ext
```diff
@@ -0,0 +1,44 @@
...
+
+# Give people an idea what to expect from this code
+#  - Is it just a test or something you stand behind?
+status stable
+
+# One line stating what the module does
+description This module implements the full 3D Dosimetry for molecular radiotherapy on multiple time points.
+
+# Space separated list of urls
+screenshoturls
```
lassoan: At least a few URLs are required

lassoan:
Thank you for your hard work and excellent contribution. We just now have to ensure that your work is easily discoverable and understandable.

Repository name is expected to Slicer<ExtensionName>. Could you rename your repository to be SlicerOpenDose4D? Also note the capitalization - Opendose is harder to read/interpret than OpenDose.

I would suggest to rework your project page (https://gitlab.com/opendose/opendose4d) to be more user friendly. Currently it mostly contains information that is only interesting for developers and there is no user documentation (tutorials, sample data sets, summary of modules and their features, etc.). Without user documentation, there is no chance that new users cannot find and start using your extension.

Please address the code review comments and complete all the items in the checklist.

OpenDose4D.s4ext
BishopWolf: do you mean outside project?

BishopWolf:
Sample datasets are inside in Resources and the user manual is in the project root

lassoan:
> Sample datasets are inside in Resources and the user manual is in the project root.

Provide links to these near the top of https://gitlab.com/opendose/opendose4d/-/blob/develop/README.md

Add a few nice screenshots.

Make sample data easy to download and use. For example, you can add a hidden module that registers custom sample data in SampleData module. This allows users to download data sets by a few clicks in Slicer. See for example here: https://github.com/SlicerRt/Sequences/tree/master/SequenceSampleData

Move developer information (Testing, CI, ...) to the very bottom of the README.md but maybe even into a separate developers.md file, to not overload users with lots of information that is not interesting for them.

> Where do I change the repository name? Is it the project name in gitlab?

Yes.

BishopWolf:
> For example, you can add a hidden module that registers custom sample data in SampleData module.

This is already in place when you hit the Reload and Test button, the whole workflow is tested, including downloading the sample data.

lassoan:
Automated tests are great, Reload&Test is very useful during development - but these are not directly relevant for users.

Providing easy access to sample data is important, as it makes it easier to follow tutorials, try various modules, troubleshoot problems (see if the problem is with the user's data), etc.

BishopWolf:
> Providing easy access to sample data is important, as it makes it easier to follow tutorials, try various modules, troubleshoot problems (see if the problem is with the user's data), etc.

This is harder for me as I can't upload data directly to the midas repository. I think I will rely on the data already provided inside the module as it is already accesible. Also it is explained in the manual.

BishopWolf:
> > Where do I change the repository name? Is it the project name in gitlab?
> 
> Yes.

I have already changed the repository name but the check still fails because of the old name, so what else shall I change?

lassoan:
> I can't upload data directly to the midas repository

You can upload your data anywhere. If you have a set of DICOM files then you can put them all in a zip file and Slicer can download and unzip them.

Can you upload arbitrary set of binary files to gitlab releases? We use github releases to host test/sample data (to avoid blowing up the git repository with lots of binary data): https://github.com/Slicer/SlicerTestingData

BishopWolf:
> You can upload your data anywhere.

I have my sample data in Resources directory, it is part of the gitlab repository

lassoan:
There are a few issues:
- [ ] The module is not packaged correctly. If I install the created package (that you can download from here: https://1drv.ms/u/s!Arm_AFxB9yqHuvEp7Ur3XyaSd9u85w?e=P0TqXb) then the module throws this error:
Traceback (most recent call last):
  File "C:/D/opendose4d_R/_CPack_Packages/win64/ZIP/28862-win-amd64-SlicerOpenDose4D-git-2020-03-20/lib/Slicer-4.11/qt-scripted-modules/Dosimetry4D.py", line 221, in setup
    self.calibration['SPECTSensitivity']['Units'], qt.Qt.MatchFixedString)
KeyError: 'SPECTSensitivity'
- [ ] User manual: Creating user manual in word is not good for many reasons. It is a large binary file. You cannot easily edit (you need to download, load it into word, edit, commit, push; instead of just making changes directly in a web browser), users cannot easily view a word file (it is relatively large, has to be downloaded, needs MS Office to view), and users cannot easily contribute (if a user contributes a small fix then you won't see in the pull request what was changed exactly). Adding a simple markdown files in the repository solves all these issues.
- [ ] The extension package is too large (20MB). What you actually need is 2-3MB python code and maybe a few resource files, but you placed 55MB extra files (sample data files?) in the package. This would increase the burden to the appstore for no good reason. All the large resource files should be downloaded dynamically. SampleData infrastructure can be used for this, it is very robust and the downloaded files are are automatically cached, so you only need to download once.
- [ ] Large binary files are stored in the source code repository. Instead, these large files should be stored in other repositories and files should be downloaded on request. See RSNAVisTutorial.py for an example how to download a DICOM folder and import its content.
- [ ] Sample data sets: The easiest way for users to get started is to add one or more sample data sets in the scene. If the input is typically DICOM then it may be easier if you import the DICOM and save the scene as an MRB file. Could you create such a scene?

BishopWolf:
Thanks for your comments:

1. this is a hidden bug, if installed from gitlab either by cloning or zipping it doesn't show. But it is a bug and will be corrected in the next push
2. Maybe referring to a google doc would be enough, but in this case user will contribute outside gitlab and that is not the intention, I will investigate the markdown way.
3. The package contains sample data (DICOM, 35MB) and a database (SQLITE, 15MB). The DICOM Data is fixed Sample Data that can follow your Sample Data suggestion. The SQLITE database is a critical part of the module and will increase in future, this does not follow the sample data approach as it is a core data in the module, what do you suggest in this case?

lassoan:
15MB database file:
- Storage/revision control: If you change it rarely then it is probably OK to keep it in the source code repository. If you change it every month or so then it is probably better to keep it as a release file (or in a separate repository) to not blow up your source code repository size.
- Installation: If users never modify it and the file compresses well then including it in the install package is probably fine. If users may modify it locally then you probably want to keep a single copy of it in the user's profile folder and only download it if the file does not exist yet.

By the way, I had a look at the content of the database file and it seems extremely inefficient and redundant to use an SQL database for this. You can use VTK data set to store arbitrary set of named arrays (scalars, vectors, strings, etc.) for a set of points or cells. You can also save metadata in field data (to store energy, half life, etc.), so one VTK file can contain all information of an isotope. If interoperability (or the ability of users to easily create/customize isotope files) is more important than performance (it may take 1-2 seconds to parse a long text file) then probably json, csv, or xml format would work much better (probably one file per isotope). These text files also can be compressed really well and since they are text files, you can track their changes in the git repository.

BishopWolf:
Progress Update
* The database has been converted to a set of JSON files
* The Manual has been refactored as MarkDown file.

Pending tasks
* Test the manual fully
* Use the Sample Data upload mechanism for the DICOM data.

BishopWolf:
> * Sample data sets: The easiest way for users to get started is to add one or more sample data sets in the scene. If the input is typically DICOM then it may be easier if you import the DICOM and save the scene as an MRB file. Could you create such a scene?

Is there an example of a testing performed in python using sample data sets? I get the point on uploading the dicom files but I still don't know how to use them from sample data module.

lassoan:
See Sequences extension to see how to register custom sample data sets and Crop volume sequence module for an example of using one of these custom data sets in module tests: https://github.com/SlicerRt/Sequences/blob/3c45c25143f6a2577851e5851747ddbb954b7459/CropVolumeSequence/CropVolumeSequence.py#L268

lassoan:
By the way, very nice progress with the database and manual.

lassoan:
@BishopWolf your extension is in great shape and while there is always room for improvements, there is nothing that should prevent adding the extension to the extension manager. Let me know if there is anything that you would like to still work on before. If not, then we can merge the pull request as is.

BishopWolf:
@lassoan Sorry for the very late response, I was full of work and didn't check the github messages so I totally missed your message until now.

I was previously debugging some instability issues with the display nodes that made slicer to crash. Fortunately that is solved.
We also nearly finished the validation prior to the publication. I expect to send it before october. 

I am currently remaking the database as some users observe some bug in the data uncertainty. After that it can be merged.

lassoan:
I merge this now so we can see any build and packaging errors on the dashboard.

BishopWolf:
> I merge this now so we can see any build and packaging errors on the dashboard.

Is there a reason why the extension doesn't show in the cdash builds? (without errors showing I don't know what happens)

jamesobutler:
Maybe you had checked the dashboard before the extension had been built?

You can use this link to see on the dashboard that there are currently configure errors
http://slicer.cdash.org/index.php?project=SlicerPreview&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=Opendose

lassoan:
It seems that gitlab requires some non-trivial configuration setting for cloning the repository when using git protocol. Probably switching to https protocol will fix the problem.

Not a problem, I'm just curious: why did you choose to use gitlab instead of github?

BishopWolf:
> Not a problem, I'm just curious: why did you choose to use gitlab instead of github?

gitlab has inside CI/CD. It has a gitlab-runner for local tests. it has a full user management and statistics. By the time I started the project github doesn't allowed private repositories.

Currently I highly depend on gitlab-runner to make a switch

BishopWolf:
> t seems that gitlab requires some non-trivial configuration setting for cloning the repository when using git protocol. Probably switching to https protocol will fix the problem.

Usually this is required only for private repositories, which it is not the case. Let me dig in the logs.

lassoan:
Thank you. I heard that github actions have changed the CI/CD landscape and now github has an edge over gitlab. Do you know what additional user control or statistics gitlab provides?

BishopWolf:
> Thank you. I heard that github actions have changed the CI/CD landscape and now github has an edge over gitlab. Do you know what additional user control or statistics gitlab provides?

For private repositories ist has full user management, you assign roles to groups of users. That was really a need when I started. After making the repository public I have no more control than protected branches

but gitlab-runner is now everything to me. I know in github you can run jenkins or circle, but having it running in my own machine allows me to gain time while testing. I put the testing to run and keep programming, when the test ends I make a commit and start a new test. When I want to commit to gitlab then I just squeeze everything.

BishopWolf:
> You can use this link to see on the dashboard that there are currently configure errors

The correct git clone line shall be `git clone git@gitlab.com:opendose/opendose3d.git` and in the error it says it attempted to clone as `git clone git://gitlab.com/opendose/OpenDose3D.git`. 

There is a chance to clone as https as `git clone https://gitlab.com/opendose/OpenDose3D`

PR #1692, opened 2020-04-09
ENH: Update ModelToModelDistance extension [master]

jcfr:
This updates the ModelToModelDistance extension to 4b4e478f9e20ecef0f66a20eb6719bd4a8c4d949.

See https://github.com/NIRALUser/3DMetricTools/compare/4b4e478f9e20ecef0f66a20eb6719bd4a8c4d949...4b4e478f9e20ecef0f66a20eb6719bd4a8c4d949 to view changes to the extension.


lassoan:
As far as I remember, ModelToModelDistance extension contains a trivial CLI (could be replaced by 4 lines of Python code) and a modified vtkImplicitPolyDataDistance filter. It would be probably better to backport the few extra options/computations to VTK proper. This feature is so simple and so commonly needed that I think we should bundle it with Slicer (in the bundled SurfaceToolbox extension).

PR #1709, opened 2020-06-05
ENH: Add TorchIO extension [master]

jcfr:
Description:
TorchIO is a Python package containing a set of tools to efficiently
read, preprocess, sample, augment, and write 3D medical images in deep
learning applications written in PyTorch, including intensity and
spatial transforms for data augmentation and preprocessing. Transforms
include typical computer vision operations such as random affine
transformations and also domain-specific ones such as simulation of
intensity artifacts due to MRI magnetic field inhomogeneity or k-space
motion artifacts.

Contributors:
Fernando Pérez-García (University College London)

### DIFF
--- a/TorchIO.s4ext
+++ b/TorchIO.s4ext
@@ -1,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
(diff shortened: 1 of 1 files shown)


lassoan:
This extension looks very interesting but I have no idea what it can do and how to use it. If this commit is only for testing if it builds then it's fine, but if we want to give a chance for users to discover and be able to use this extension then we need to complete the extension checklist.

PR #1716, opened 2020-06-15
ENH: Adds SlicerMorph as a dependency

sshanshans:
ENH: add SlicerMorph as a dependency to Auto3dgm extension

### DIFF
--- a/Auto3dgm.s4ext
+++ b/Auto3dgm.s4ext
@@ -6,7 +6,7 @@ scmrevision master
 # list dependencies
 # - These should be names of other modules that have .s4ext files
 # - The dependencies will be built first
-depends NA
+depends SlicerMorph
 
 # Inner build directory (default is ".")
 build_subdirectory .


sshanshans:
Hi James, thanks so much for pointing out the circular dependency issue! Somehow it escaped my radar of fallacies.

To answer your question, this is not a true dependency, and we would like to let users of Auto3dgm to also download SlicerMorph. Any advice on how to achieve this will be greatly appreciated!

jamesobutler:
So it sounds like Auto3dgm just wants to recommend that another extension (SlicerMorph) would be useful to have for related work.

Is there a reason why Auto3dgm modules are not part of the SlicerMorph extension if they are that closely related to various workflows?

Also has there been consideration for using a [Slicer Custom App](https://github.com/KitwareMedical/SlicerCustomAppTemplate) to prepackage a set of Slicer extensions for a certain set of users with custom branding?

@lassoan @jcfr Has there been any other extensions that have been so closely related that one would require the download of the other even though they aren’t a true dependency and just recommended for the type of work that a user is likely doing?

pieper:
Hi @jamesobutler - yes, SlicerMorph and Auto3dgm are part of the same [overall development project](https://slicermorph.github.io/) just developed independently and we do currently provide "batteries included" apps for end users, but also want them as extensions for easy testing.

lassoan:
There have been examples of complex dependencies in SlicerIGT, OpenIGTLink, IGSIO, Sequences, PerkTutor extensions. You can always resolve these by creating an "infrastructure" extension that contains common parts that both extensions depend on.

However, I'm not sure if SlicerMorph/Auto3dgm is a real dependency. If you just want to make it simple for users to install ineresting modules then I would recommend to add a "SlicerMorph Home" or "SlicerMorph Advisor"module, which would advise the user about what extensions he may consider to install.

Here is an example of how to download and install extension from a script: https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository#Download_and_install_extension

sshanshans:
Hi @jamesobutler, @pieper and @lassoan -- thanks so much for your help! I'm very excited to try out the different options that are suggested here. I will talk to my team about which way to proceed next. But for now, I will close this pull request, as the circular dependency will be greatly troublesome. Thanks again for all your help!

PR #1720, opened 2020-06-30
SlicerCaseIterator: added missing dependency to SlicerDevelopmentToolbox

che85:
Added missing dependency to SlicerDevelopmentToolbox

for reference: https://github.com/JoostJM/SlicerCaseIterator/issues/15


che85:
Thanks. Good point!

for reference: https://github.com/JoostJM/SlicerCaseIterator/pull/17

che85:
I think this is ready to merge. Thank you

PR #1728, opened 2020-08-21
Add CollageRadiomics extension.

nathanhillyer:
- [ ] Repository name is Slicer+ExtensionName. **We are keeping a single repo, and requesting an exception to this guideline**
- [x] Repository is associated with `3d-slicer-extension` GitHub topic so that it is listed [here](https://github.com/topics/3d-slicer-extension). To edit topics, click "Manage topics" on your repository's main page, on the left, near the top (just below the repository description). To learn more about topics, read https://help.github.com/en/articles/about-topics
  - [x] Extension name
  - [x] License: We suggest you use a permissive license that includes patent and contribution clauses.  This will help protect developers and ensure the code remains freely available.  We suggest you use the [Slicer License](https://github.com/Slicer/Slicer/blob/master/License.txt) or the [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0). Always mention in your README file the license you have chosen.  If you choose a different license, explain why to the extension maintainers. Depending on the license we may not be able to host your work. Read [here](https://opensource.guide/legal/#which-open-source-license-is-appropriate-for-my-project) to learn more about licenses.

### DIFF
--- a/CollageRadiomics.s4ext
+++ b/CollageRadiomics.s4ext
@@ -1,0 +1,23 @@
+scm git
+scmurl https://github.com/radxtools/collageradiomics.git
+scmrevision master
+
+depends NA
+
+build_subdirectory .
+
+homepage https://github.com/radxtools/collageradiomics/tree/master/slicer/collageradiomics
+
+contributors Pallavi Tiwari (BriC), Satish Viswanath (BriC), Rob Toth (Toth Technology), Nathan Hillyer (Toth Technology)
+
+category Informatics
+
+iconurl https://github.com/radxtools/collageradiomics/raw/master/slicer/collageradiomics/CollageRadiomicsSlicer/Resources/Icons/CollageRadiomicsSlicer.png
+
+status beta
+
+description The CollageRadiomics extension provides a 3D Slicer interface to the collageradiomics library. collageradiomics is an open-source python package for capturing subtle anisotropic differences in disease pathologies by measuring entropy of co-occurrences of voxel-level gradient orientations on imaging computed within a local neighborhood.
+
+screenshoturls https://github.com/radxtools/collageradiomics/raw/master/slicer/collageradiomics/CollageRadiomicsSlicer/Screenshot.jpg
+
+enabled 1


nathanhillyer:
> Install time may even exceed maximum module startup time, preventing loading of your module the first time you start the application after the extension installation.

Do you have any suggestions for working around this problem? I'm now realizing our pip packages are cached on our development boxes, hence we never ran into this issue. But I'm seeing this problem on an installation on a fresh computer. Are there any better patterns to follow or are we the first extension that uses python dependencies like this?

Thanks--

fedorov:
In SlicerRadiomics we [install pyradiomics and its prerequisites](https://github.com/Radiomics/SlicerRadiomics/blob/master/SuperBuild/External_python-pyradiomics.cmake#L63-L70) as part of the superbuild process. Isn't this something that can be done here?

nathanhillyer:
@fedorov Great idea. Not familiar with the superbuild process but we'll do some research. Do you have any suggestions for a starting point?

fedorov:
For me the starting point is usually a piece of code that does something similar to what I want to do, and in this case it seems like SlicerRadiomics might be that piece. You can then follow up on the [Slicer Discourse](https://discourse.slicer.org) if you have specific questions. @lassoan may give a better pointers.

nathanhillyer:
Thanks. I've started on the work here, following the pyradiomics patterns:
https://github.com/radxtools/collageradiomics/commit/f17f291ec99c66d86ab80b2332da90bc6ffeaa1a

Does anyone know a good way to test these super builds locally?

lassoan:
If you already have wheels on PyPI then it is easier to pip install the package, as it is done here: https://github.com/ToothAndClaw/SlicerAuto3dgm/blob/13cd3d98607c4ecddb45c7bdc02cb76569dd89bc/Auto3dgm/Auto3dgm.py#L72-L77

Note that this is in the module widget's constructor, so it is not executed until the user switches to the module and a popup is displayed so the users knows what is going to happen and it may take time.

nathanhillyer:
That looks good, as we do have wheels on PyPI. As always, we very much appreciate the help.

fedorov:
@lassoan for my curiosity - what are the disadvantages to installing those via pip during superbuild, as we do in SlicerRadiomics?

lassoan:
If you build a python package in the Slicer extension then it is one more thing that you need to maintain. If you already have your package in PyPI then you can get that from Slicer instead of building it again.

Another thing is that if the Python package depends on other packages and you want to bundle all those in your extension package then its size can become quite big.

I'm not sure what happens if a bundled Python package's version conflicts with another extension's requirements. There may be differences in how a package can be uninstalled and a different version reinstalled - @jcfr may know this.

The only advantage of bundling required Python packages with the extension is that it is more self-contained: after you downloaded and installed the extension you don't need internet connection for the first use of the extension.

lassoan:
@nathanhillyer Have you finished with all the updates that were requested and you have planned to do?

nathanhillyer:
@lassoan We had some things come up and have been working on integrating with other software projects the past couple weeks that were scheduled as higher priority, and we'll circle back to this one soon. Will ping when we get back to this!

fedorov:
Someone brought it to my attention that CoLIAge is covered by this patent: https://patents.google.com/patent/US9483822B2/en.

How is this related to the implementation in this extension? I did not find any mention of the patent on the front page of https://github.com/radxtools/collageradiomics.

I believe Slicer license is very strict that any code contributed to the application should not be subject to any patent protection. I do not know if/how this applies to extensions. FWIF I did not see "patent" mentioned in https://www.slicer.org/wiki/Documentation/Nightly/Developers/FAQ/Extensions.

pieper:
Thanks for raising this @fedorov, I was not aware of it either.

I agree it is very important to disclose any patents that may influence a user's ability to make use of the code.  And to the extent possible it is important to be explicit about terms of such use.  For example some people state that patents can be used royalty-free for research purposes but that commercial users would be expected to pay for a license.  It's only fair to make these kinds of conditions clear up front so that people can decide whether to make use of this extension or not in their work.

If this code were being contributed to the Slicer core, then the [patent terms would come into effect](https://github.com/Slicer/Slicer/blob/master/License.txt#L62-L71).  But since this is an extension and is made available under the BSD license the patent terms are not proscribed and should be explicitly called out.

I'll note that by making the extension available the authors are in effect encouraging people to use the patented material without claiming your IP.  You may wish to consult your lawyers about the impact on future ability to assert any control over that IP.

nathanhillyer:
Thanks @pieper. The extension is being developed under the direction of the patent writers and I'll bring up these concerns.

lassoan:
Agree, we don't impose any particular license on extensions (only on contributions to Slicer core).

However, we have been asking developers of all extension that used restrictive license (for example, GPL) to mention it explicitly in their extension description. We haven't asked developers about known related patents, but of course that should be disclosed in the extension description, too. I've added it to the [extension checklist](https://github.com/Slicer/ExtensionsIndex/blob/master/.github/PULL_REQUEST_TEMPLATE.md) now to make sure we don't forget about this.

lassoan:
I think allowing businesses to distribute their offerings as Slicer extensions is good for everyone, as it allows fair comparison and competition of various tools. The only thing we need to pay attention to is not to mislead users to think that a certain tool is free (in every sense), while it is actually not. If more restrictively licensed/patented extensions show up then we may add a filtering option in the extension manager that makes is easier to users to find free solutions.

fedorov:
@lassoan I agree. Also note that not disclosing the patent, while sharing code under a non-restrictive license may be a deterrent for the users who know about the patent, but do not know how to reconcile it with the license. When I learned about this today, I actually could not make sense of how the BSD license and patent co-exist. @pieper explanation helped understand that part somewhat! :-)

lassoan:
I've been reviewing open pull request and noticed that this extension is stuck in review phase. Is there still an intention to make the CollageRadiomics available as a Slicer extension?

lassoan:
I've been reviewing open pull request and noticed that this extension is stuck in review phase. Is there still an intention to make the CollageRadiomics available as a Slicer extension?

lassoan:
Closing due to no response. Feel free to reopen if ready to submit the extension.

PR #1742, opened 2020-10-21
Update SlicerDcm2nii.s4ext

pieper:
Add dedicated 4.11 branch so that master can be updated for vtk 9

### DIFF
--- a/SlicerDcm2nii.s4ext
+++ b/SlicerDcm2nii.s4ext
@@ -1,4 +1,4 @@
-#
+s#
 # First token of each non-comment line is the keyword and the rest of the line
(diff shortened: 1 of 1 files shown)


SlicerDcm2nii.s4ext
```diff
@@ -1,4 +1,4 @@
-#
+s#
```
jamesobutler, line 1: An unintentional letter was added here

lassoan:
Committed along with @jamesobutler's fix in https://github.com/Slicer/ExtensionsIndex/commit/4e044ce28b733d94e5a5ef057d80424b75b8c6db

PR #1757, opened 2021-04-23
Renaming extension to SlicerPinholeCameras

adamrankin:


pieper (approved):
LGTM

pieper:
There's a CI failure but it seems unrelated.  I believe we can override the commit check here.  @lassoan @jcfr do you agree?

jcfr:
Thanks for the ping. Will have a look in 20mins or so.

jcfr:
I rebased the branch against current master and forced push

jcfr (approved):

PR #1767, opened 2021-05-27
Update TCIABrowser iconurl

jamesobutler:
This updates the TCIABrowser iconurl to use the one in the repo.


lassoan (approved):
thank you!

PR #1770, opened 2021-06-27
Add the new Similarity Subgroups Extension to the Index

sebastianandress:
This method not only finds surface deviations of a model, but also shows multiple high accuracy zones (with deviations less than a definable threshold) of a potentially corrupted model. We used it for the validation of 32 models that were used for surgery planning.
The advantage to standard validation methods is, that the surgeon can select a subgroup that fits his region of interest, and can be sure the model is save to use, even when it may be corrupt at other regions.

Thank you for contributing to 3D Slicer!

- **To add a new extension with this pull request**: Please keep content of "TODO list for submitting a new extension" section and put an 'x' character in the brackets for each todo item to indicate that you have accomplished that prerequisite.

- **To update an existing extension with this pull request**: Please delete all text in this template and just describe which extension is updated and optionally tell us in a sentence what has been changed. To make extension updates easier in the future (so that you don't have to submit a new pull request after each change in your extension), you may consider replacing specific git hash in your s4ext file by a branch name (for example: `master` or `release` for nightly version; `(majorVersion).(minorVersion)` such as `4.10` for stable Slicer version).

# TODO list for submitting a new extension

[To make sure users can find your extension, understand what it is intended for and how to use it, please complete the checklist below. You do not need to complete all the item by the time you submit the pull request, but most likely the changes will only be merged if all the tasks are done. See more information about the submission process here: https://slicer.readthedocs.io/en/latest/developer_guide/extensions.html]

- [x] If source code license is more restrictive for users than BSD, MIT, Apache, or 3D Slicer license then the name of the used license must be mentioned in the extension description.
- [x] Extension icon URL is correct (do not use the icon's webpage but the raw data download URL that you get from the download button - it should look something like this: https://github.com/user/project/raw/main/path/to/icon.png)
  - [x] Extension name
  - [x] License: We suggest you use a permissive license that includes patent and contribution clauses.  This will help protect developers and ensure the code remains freely available.  We suggest you use the [Slicer License](https://github.com/Slicer/Slicer/blob/master/License.txt) or the [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0). Always mention in your README file the license you have chosen.  If you choose a different license, explain why to the extension maintainers. Depending on the license we may not be able to host your work. Read [here](https://opensource.guide/legal/#which-open-source-license-is-appropriate-for-my-project) to learn more about licenses.

[If you have any questions or comments then please describe them here.]

### DIFF
--- a/SimilaritySubgroups.s4ext
+++ b/SimilaritySubgroups.s4ext
@@ -1,0 +1,42 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl git://github.com/sebastianandress/Slicer-SimilaritySubgroups.git
+scmrevision master
+
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends NA
+
+# Inner build directory (default is ".")
+build_subdirectory .
+
+# homepage
+homepage https://github.com/sebastianandress/Slicer-SimilaritySubgroups
+
+contributors Sebastian Andress (LMU Munich)
+
+# Match category in the xml description of the module (where it shows up in Modules menu)
+category Segmentation
+
+# url to icon (png, size 128x128 pixels)
+iconurl https://github.com/sebastianandress/Slicer-SimilaritySubgroups/raw/master/SimilaritySubgroups.png
+
+# Give people an idea what to expect from this code
+#  - Is it just a test or something you stand behind?
+status Beta
+
+# One line stating what the module does
+A polydata/model validation method that finds multiple high accuracy similarity zones on a source model deviating less than a definable threshold from a target model.
+
+# Space separated list of urls
+screenshoturls https://github.com/sebastianandress/Slicer-SurfaceWrapSolidify/raw/master/Resources/screenshot1.png
+
+# 0 or 1: Define if the extension should be enabled after its installation.
+enabled 1
--- a/SurfaceFragmentsRegistration.s4ext
+++ b/SurfaceFragmentsRegistration.s4ext
@@ -6,7 +6,7 @@
 
 # This is source code manager (i.e. svn)
 scm git
-scmurl git://github.com/sebastianandress/Slicer-SimilaritySubgroups.git
+scmurl git://github.com/sebastianandress/Slicer-SurfaceFragmentsRegistration.git
 scmrevision master
 
 # list dependencies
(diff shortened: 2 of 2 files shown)


lassoan:
Thanks a lot for your contribution, this is very interesting and useful work. The extension is implemented very nicely, including a small automatic test; documentation also very informative. A couple of things to address:

High priority:
- [ ] Installation of `sklearn` is missing. You can install it in the widget's `setup` method. Very important: you must remove all `from sklearn...` calls from the file's scope, because Slicer will not be able to load your module at all (and the module will not have a chance to install sklearn). These imports should be placed in the methods where they are used.
- [ ] The extension's name does not give any hint what the extension may be usable for. Something like "Piecewise rigid registration", "Model fragment registration", "Surface zone matching", etc. would be much better.
- [ ] Follow the current scripted module template (generate a new module using latest Extension wizard module and copy-paste methods). The most painful omission in the current implementation is that none of the parameters are stored in the scene. This means that after the user carefully sets 8-10 parameters and saves the scene; next time the user loads the scene, all those parameter settings are gone!

Medium priority:
- [ ] It is not clear what the module can be used for. Please describe a few applications. It is also not clear what the output of the algorithm is. Colored scalars?
- [ ] Add tooltips to every sliders and checkboxes
- [ ] Add a markups place widget next to the source/landmark node selectors. Without place widgets it is very difficult to place points. See for example [Curved planar reformat module in Sandbox extension](https://github.com/PerkLab/SlicerSandbox/tree/master/CurvedPlanarReformat).
- [ ] Add a display section to the end (below the Apply button) that can show the various scalar overlays. It can be just a few buttons that make the corresponding scalar value shown for the model (maybe they could set color map, transparency, add scalar bar, etc. - whatever you usually do when you review the results).
- [ ] Move all the checkboxes to the right side of the label to conform with Slicer GUI design conventions 
- [ ] Move all the parameters that are not required to be tuned for _every_ case to an "Advanced" section (collapsed by default).
- [ ] Add a more realistic sample data set. Registering a sphere with a double-sphere does not really show how this module can be used (on what kind of data). You can simulate a fracture using segment editor, by splitting a segment into two segments using Scissors effect, exporting them to models, then move one of the models and harden the transform. You can then upload these models into your extension's repository (as a release asset), and register these models as sample data. See example [here](https://github.com/SlicerMorph/SlicerMorph/blob/13fb70e38457ccd00e7e51a125d2680b89c7f9ea/SlicerMorphSampleData/SlicerMorphSampleData.py#L44-L53) for adding a model and markups node as a sample data set.

Low priority:
- [ ] Significant portion of the code is not used (`create*Deformation` methods). Is this intentional? Are they going to be used in future versions? At least they could be used in some automated testing.
- [ ] Provide alignment transforms as output. This would hugely increase the usefulness of the module for a wide range of problems.

sebastianandress:
Dear Andras,
thanks again for the high quality feedback. I incorporated all the suggested changes and renamed the module to 'Surface Fragments Registration'.

High priority:
- [x] Removed methods using sklearn, as they were deprecated.
- [x] Renamed the extension.
- [x] Used the new template, now all parameters are stored in a parameter node, and the ui is built up from an ui file.

Medium priority:
- [x] Hopefully the outputs are more clear now, as there is a new output section that allows for spinning through the resulting registration transforms and according surface deviation scalars.
- [x] Added tooltips to all parameters.
- [ ] As the landmarks registration/selection is now in the advanced section, I did not add a placing widget to keep the module as simple as possible.
- [x] An output section is added.
- [x] UI changed a lot, hopefully the module is now conform to the GUI design conventions.
- [x] An advanced section is added.
- [x] A more realistic example with a artificially fractured iliac crest is added.

Low priority:
- [x] Deprecated code is removed.
- [x] Transforms as output are provided now.

lassoan:
Thank you for these improvements. I'll not be able to test this before early next week. Maybe @pieper or others can give it a try in the meantime and give you further feedback or merge the extension.

sebastianandress:
I updated the extension again to make it more usable by removing the `Initialisation Candidate Radius` parameter and therefore using the `Minimal Fragment Size`, as these values correspond to each other in a certain way.

pieper:
I haven't had a chance to try the code, but a quick read-through looks good to me.  Thanks for contributing this!  I have no objection to merging this so people can easily test it from the extension manager.

sebastianandress:
Are there any news regarding this pull request? Thank you

pieper:
Did you want to have another look @lassoan ?

lassoan:
It should be mostly OK, so let's merge it now so that we can test it more easily.

PR #1776, opened 2021-07-01
Update monai label

diazandr3s:
Thank you for contributing to 3D Slicer!

Modify special characters


lassoan:
It seems that this PR has been already merged.

PR #1797, opened 2021-11-08
migrate SlicerANTs organization

simonoxen:
I migrated SlicerANTs repo to the netstim organization


lassoan (approved):
Perfect, thank you. I see that you have correctly updated the organization name in CMakeLists.txt and other source file, too.

PR #1801, opened 2021-12-21
remove mpReview

fedorov:


jcfr:
Thanks for taking care of this.

Could you add an entry to https://github.com/Slicer/ExtensionsIndex/blob/master/ARCHIVE/README.md ?

jcfr:
Based on the check, the following error is reported:

```
mpReview extension is not found. It is required by extension: SliceTracker.
```

Since following https://github.com/Slicer/ExtensionsIndex/commit/d5b53aa3ff3d71c2c12a39a692a0fde7ed2a917f, `SliceTracker` was updated to include a dependency on `SegmentEditorExtraEffects`, may it could be as simple as updating `SliceTracker` to remove dependency on the legacy editor and relying only on the segment editor ?

fedorov:
Unfortunately, SliceTracker is relying on some other components in mpReview, and will need to be removed as well.

fedorov:
@tokjun FYI - I don't know if you are relying on SliceTracker in the nightly build, but it will not be available anymore.

PR #1816, opened 2022-04-14
Build is fixed on Linux and Windows, re-enable for build testing on Mac

adamrankin:


jamesobutler:
https://slicer.cdash.org/index.php?project=SlicerPreview&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=SlicerOpenCV

It appears the Slicer factory Linux machine has failed to build SlicerOpenCV. 

For macOS it’s interesting that the 2 build machines had different results.
- Build failed on factory-south-macOS which is macOS 10.13, Xcode 10.1(?)
- Build succeeded on computron-macOS which is macOS 10.15, Xcode 11.7

PR #1818, opened 2022-04-22
Update XNATSlicer tag with Python 3 support

jamesobutler:
XNAT needs to use the latest version of the extension with Python 3 support to have a chance at loading successfully in Slicer.


lassoan (approved):
Thank you!

PR #1825, opened 2022-05-11
ENH: Remove unnecessary dependencies

smrolfe:
Sandbox, SlicerIGT, RawImageGuess, SlicerDcm2nii, and SurfaceWrapSolidify are not true dependencies of the SlicerMorph extension, so it's not necessary to install before use. We plan to replace this default installation with an option in SlicerMorph to install recommended extensions.

### DIFF
--- a/SlicerMorph.s4ext
+++ b/SlicerMorph.s4ext
@@ -12,7 +12,7 @@ scmrevision master
 # list dependencies
 # - These should be names of other modules that have .s4ext files
 # - The dependencies will be built first
-depends     SegmentEditorExtraEffects Sandbox SlicerIGT RawImageGuess SlicerDcm2nii SurfaceWrapSolidify
+depends     SegmentEditorExtraEffects
 
 # Inner build directory (default is .)
 build_subdirectory .


SlicerMorph.s4ext
```diff
@@ -12,7 +12,7 @@ scmrevision master
 # list dependencies
 # - These should be names of other modules that have .s4ext files
 # - The dependencies will be built first
-depends     SegmentEditorExtraEffects Sandbox SlicerIGT RawImageGuess SlicerDcm2nii SurfaceWrapSolidify
+depends     SegmentEditorExtraEffects
```
jamesobutler, line 15: Which SlicerMorph module has a dependency on the SegmentEditorExtraEffects extension?

smrolfe: SegmentEditorExtraEffects is still required for our tutorials/workshops, though also not a true dependency of the modules.

SlicerMorph.s4ext
```diff
@@ -12,7 +12,7 @@ scmrevision master
 # list dependencies
 # - These should be names of other modules that have .s4ext files
 # - The dependencies will be built first
-depends     SegmentEditorExtraEffects Sandbox SlicerIGT RawImageGuess SlicerDcm2nii SurfaceWrapSolidify
```
jamesobutler, line 15: It appears the SegmentEndocranium does have a dependency on SurfaceWrapSolidify. Are you planning to remove this dependency as part of the removal here?
https://github.com/SlicerMorph/SlicerMorph/blob/2be23fa5adf86d57da38a5c799e095c92d4ef689/SegmentEndocranium/SegmentEndocranium.py#L248-L249

SlicerMorph.s4ext
jamesobutler: > We plan to replace this default installation with an option in SlicerMorph to install recommended extensions.

I figured this is what you would do with SegmentEditorExtraEffects as it is another recommended extension to install, but not a true dependency of SlicerMorph modules.

SlicerMorph.s4ext
smrolfe: Yes, this will be changed to install SurfaceWrapSolidify from this module.

jamesobutler: Would you maintain SurfaceWrapSolidfy dependency at https://github.com/SlicerMorph/SlicerMorph/blob/2be23fa5adf86d57da38a5c799e095c92d4ef689/CMakeLists.txt#L13? 

When including SlicerMorph modules into a Slicer custom app dependent extensions will need to be stated somewhere. Slicer custom apps don't generally install extensions from the ExtensionsManager (from Slicer factory online store) because there could be incompatibilities between the factory build configuration and the Slicer custom app configuration. Therefore it would need to be known to build SurfaceWrapSolidify as part of the Slicer custom app build process so SegmentEndocranium could continue to work.

smrolfe: The dependencies would be removed from the EXTENSION_DEPENDS field of the CMake file. I can add instructions in the extension documentation for including SurfaceWrapSolidify when building a custom app, or let me know if you have another suggestion for where to include this.

jcfr: Listing the modules and related extensions (aka soft dependencies) in the documentation would be ideal.

smrolfe: Thanks @jcfr I will add this to the extension documentation.

SlicerMorph.s4ext
smrolfe: While the other extensions are helpful for some, almost all SlicerMorph users will need SegmentEditorExtraEffects, so we'd like to retain it in the default distribution. This single dependency should not have a large impact on the build order, which was our main [motivation for removing the others](https://github.com/SlicerMorph/SlicerMorph/issues/195).

smrolfe:
Checking in on the status of this pull request, are there any issues left to address? The corresponding changes to the SlicerMorph repo are in [here](https://github.com/SlicerMorph/SlicerMorph/pull/211).

jcfr:
Consider also merging https://github.com/SlicerMorph/SlicerMorph/pull/211

jcfr:
@smrolfe Could you also backport this change into the 5.0 branch ? :pray: 

See https://github.com/Slicer/ExtensionsIndex/blob/5.0/SlicerMorph.s4ext#L9-L10

smrolfe:
Thanks @jcfr I have submitted a pull request for the 5.0 branch here: https://github.com/Slicer/ExtensionsIndex/pull/1829

PR #1863, opened 2022-07-17
ENH: Update SlicerColocZStats extension

ChenXiang96:
By using Numpy indexing to threshold all selected volumes and identify intersections between different volumes, the computational speed of the extension is greatly improved.

### DIFF
--- a/SlicerColocZStats.s4ext
+++ b/SlicerColocZStats.s4ext
@@ -6,8 +6,8 @@
 
 # This is source code manager (i.e. svn)
 scm git
-scmurl https://github.com/ChenXiang96/SlicerColoc-Z-Stats.git
-scmrevision main
+scmurl https://github.com/ChenXiang96/SlicerColoc-Z-Stats
+scmrevision 84df25ed0bf52d0e9bdb16557c8e2f72d173d164
 
 # list dependencies
 # - These should be names of other modules that have .s4ext files


jamesobutler (commented):
@ChenXiang96 This change isn’t necessary if you’re trying to make this latest performance improvement available to users. Since the “main” branch is specified, the extension will always be built with the latest code on the branch and be distributed to users.

ChenXiang96:
Hello @jamesobutler ,
   Thank you for your kind reply.
   I've update my code of both the "main" and "5.0" branch 3 days ago.  
   However, after I downloaded the stable version of slicer and  installed the extension from 'Extension manager', I found that the extension is still the same as it was before I updated it, what is the reason for this and what should I do?
   Thank you so much!

lassoan:
You can always check the status of your extension (what version was built, on which platforms, if there were any build and test warnings or errors, etc.) on the [dashboard](https://slicer.readthedocs.io/en/latest/developer_guide/contributing.html?highlight=cdash#nightly-tests).

For example:
https://slicer.cdash.org/index.php?project=SlicerStable&date=2022-07-17&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=Coloc

As you can see, your extension was built and packaged successfully for latest Slicer Stable Release, which is 5.0.3 (30893). On the Slicer download site the link still points to 5.0.2 (30822) due to some delays with code signing of the 5.0.3 package (as described [here](https://discourse.slicer.org/t/slicer-5-0-x-stable-downloads/24278/8?u=lassoan)).

The new Windows Slicer-5.0.3 stable release should be uploaded today, which will fix your problem. Until then you can download the unsigned release from [here](https://slicer-packages.kitware.com/#item/62c7fb6aaa08d161a319b428).

I'm closing this issue because the git hash should not be updated in the s4ext file and it would actually prevent automatic updates of your extension.

PR #1882, opened 2022-08-18
BUG: Use HTTPS SlicerMarkupConstraints homepage

allemangD:
Markup Constraints homepage should use HTTPS. The change is reflected in the CMakeLists here:

https://github.com/KitwareMedical/SlicerMarkupConstraints/commit/479117917e29a8e99299740fb558996928070dbf

### DIFF
--- a/SlicerMarkupConstraints.s4ext
+++ b/SlicerMarkupConstraints.s4ext
@@ -18,7 +18,7 @@ depends NA
 build_subdirectory .
 
 # homepage
-homepage http://github.com/KitwareMedical/SlicerMarkupConstraints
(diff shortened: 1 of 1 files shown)


jamesobutler (approved):

jamesobutler:
Because the branch originated from https://github.com/Slicer/ExtensionsIndex rather than a personal fork? Only non-typical use I see here

jcfr:
That is a good point, that said the other PR where we observed seems to have been created from @allemangD fork. See https://github.com/Slicer/ExtensionsIndex/pull/1878 

We can re-evaluate in the future when (and if) the issue happen again ....

PR #1896, opened 2022-11-18
5.2: Update SlicerMorph to a fixed version

jcfr:
This commit updates to a fixed version to ensure build associated with the current stable works as we are updating SlicerMorph to relies on features available only in the preview build of Slicer.


jcfr:
cc: @smrolfe

PR #1913, opened 2023-02-22
Add PyTorch extension as a dependency

smrolfe:
Update MEMOS to add PyTorch extension as a dependency.


jamesobutler:
@smrolfe This also needs to be backported to the https://github.com/Slicer/ExtensionsIndex/tree/5.2 branch?

PR #1920, opened 2023-03-12
Add new extension AI Assisted BUS Diagnosis for 3D Slicer

ZhyBrian:
# New extension

- [x] LICENSE.txt is present in the repository root. MIT  (https://choosealicense.com/licenses/mit/) or Apache (https://choosealicense.com/licenses/apache-2.0/) license is recommended. If source code license is more restrictive for users than MIT, BSD, Apache, or 3D Slicer license then the name of the used license must be mentioned in the extension description.
- [x] Extension URL and revision (scmurl, scmrevision) is correct, consider using a branch name (main, release, ...) instead of a specific git has to avoid re-submitting pull request whenever the extension is updated
  - [x] Extension name
  - [x] License: We suggest you use a permissive license that includes patent and contribution clauses.  This will help protect developers and ensure the code remains freely available.  We suggest you use the [Slicer License](https://github.com/Slicer/Slicer/blob/main/License.txt) or the [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0). Always mention in your README file the license you have chosen.  If you choose a different license, explain why to the extension maintainers. Depending on the license we may not be able to host your work. Read [here](https://opensource.guide/legal/#which-open-source-license-is-appropriate-for-my-project) to learn more about licenses.


pieper:
From a quick look this looks good 👍 

A concern is that BUS is not familiar to me so I'd suggest a name like SlicerAIBreastUS or SlicerBreastDiagnosisAI.

ZhyBrian:
Hello, thank you for your quick response!
Your concern does make sense. I didn't realize that the meaning of BUS is not familiar to most people. I prefer to change the name to SlicerAIBreastUSDiagnosis, would you think that is too long or just ok?

pieper:
Shorter names are usually better but a clear descriptive name is also nice, so what you proposed makes sense to me.

It's great to see this contribution and say hello to Xiaojun for me. 👋

ZhyBrian:
OK. And surely I'll convey your regards to him!

lassoan:
A few comments on the naming:
- The `Slicer` prefix is only used for the repository name, it is not part of the extension name.
- `Diagnostic` has a very specific meaning in medical device regulations (i.e., suitable for making diagnosis), so I would not recommend to include the `Diagnosis` word in the extension name.
- Meaning of `US` is not obvious without context.
- We cannot "diagnose the breast", it has to be more specific - diagnose tumor, lesion, etc.

A few name suggestions:
- BreastUltrasoundAnalysis
- BreastUltrasoundLesionDetector
- BreastUltrasoundLesionSegmenter
- BreastUltrasoundTumorSegmenter

Please use SlicerPyTorch extension to install pytorch in Slicer, as all the other extensions do that use PyTorch. We have learned that installing PyTorch for non-developers is very challenging, there are lots of pitfalls (install/don't install CUDA, which version; what to do if the computer can technically use CUDA but the GPU is not powerful enough for PyTorch; what to do if PyTorch installation fails, what to do if CUDA version bundled with PyTorch is not compatible with the driver version, etc.). Even if just 5% of users have these problems it means that dozens of users come and complain (and hundreds of users just silently give up and conclude that your extension is broken). You can use [TotalSegmentator](https://github.com/lassoan/SlicerTotalSegmentator) as an example fo how to use SlicerPyTorch.

lassoan:
If possible, avoid using OpenCV for image processing. OpenCV is very powerful, contains lots of features, but it is just too big, not focused enough, relies on 50+ external components and its packaging and building procedure is a mess. Several percent of users will have trouble getting opencv-python installed on their computers due to build or third-party library incompatibility issues. Have a look at their [issue tracker](https://github.com/opencv/opencv-python/issues). You don't want to spend your time supporting users with diagnosing/fixing/reporting all the OpenCV build&installation issues (but if you don't do that then users will think that your extension and Slicer are garbage).

Since you only use basic image processing features, you don't need OpenCV. Instead, you can use SimpleITK (already bundled with Slicer) or ITK-Python. They have all the features that you use in your extension. Or you can use VTK (it is bundled with Slicer), which has at least most things that you use in your extension. You could also look for any smaller, more focused image processing packages that are not likely to cause problems.

If you do want to use OpenCV then please try if installing SlicerOpenCV extension is sufficient. We bundle a core part of OpenCV in this extension and it is less likely to interfere with Slicer and other Python packages than the opencv-python distribution.

ZhyBrian:
Thank you for your detailed suggestions!
I'll change a more reasonable name later. And for PyTorch and OpenCV problems, I will deal with them in the next few days. Not depending on OpenCV seems not a big problem, I will try to use SimpleITK to replace OpenCV.

lassoan:
We need to finalize the extension and repository name before merging the pull request. The other fixes can wait until public announcement and starting advertising of the extension.

ZhyBrian:
ok. I'll change the repository name to SlicerBreastUltrasoundAnalysis, and the extension name BreastUltrasoundAnalysis. Sorry for no responding cause we were just taking a rest at that time.

ZhyBrian:
Hi Andras,

Up to now I have done all the changes including:
- The repository name is changed to SlicerBreastUltrasoundAnalysis and the extension name is changed to BreastUltrasoundAnalysis.
- All OpenCV implementations in the source code have been replaced with SimpleITK.
- The SlicerPyTorch extension has been used to install PyTorch, and as a result, the 'depends' in the .s4ext file and CMakeList.txt have been updated. It's ok to use this module by simply installing pytorch of cpu version. It will take only about 1s to segment one ultrasound slice(which is often used in practice) without GPU.
- A sample data has been added, and users can download it by simply clicking the 'Download and Show Sample Data' button in this module.

Currently all items in the checklists are done. I wonder what are the following procedures and could you please give me an approximation that how long it will take to publish this extexsion in `Extensions Manager` if everything goes right?

Thank you in advance for your help and have a good day!

Yi

ZhyBrian:
Apologies for the disturbance. I'm checking on the status of this pull request and would appreciate any updates.  It has been three weeks since my previous comment, and I have not received any feedback during this period. If there are any further issues with this extension, please feel free to bring them up. I would be happy to make timely corrections.

pieper:
Thanks for the updates and the reminder @ZhyBrian - please give it a try when the build cycle completes to confirm that the installed version works as you expect.

ZhyBrian:
Great! Once the module is available on the Extension Manager, I will try it out. Thanks!

PR #1963, opened 2023-07-31
BUG: Add the `UKFTractography` dependency to the `SlicerDMRI` extension

jhlegarreta:
Add the `UKFTractography` dependency to the `SlicerDMRI` extension, following the dependenciest listed in the latter's top-level `CMakeLists.txt` file:
https://github.com/SlicerDMRI/SlicerDMRI/blob/610533480896d2884cb9ab0671c88874d237247d/CMakeLists.txt#L14

### DIFF
--- a/SlicerDMRI.s4ext
+++ b/SlicerDMRI.s4ext
@@ -2,7 +2,7 @@ scm git
 scmurl https://github.com/SlicerDMRI/SlicerDMRI.git
 scmrevision master
 
-depends NA
+depends UKFTractography
 
 build_subdirectory inner-build
 


jhlegarreta:
@pieper @zhangfanmark @ljod please check which of the two is not accurate, whether this was OK but `SlicerDMRI`'s `CMakeLists.txt` is listing the dependency unnecessarily. Thanks.

ljod:
@jhlegarreta We install it as part of SlicerDMRI because we want to always distribute it with SlicerDMRI. I vaguely recall a circular dependency issue where UKFTractography requires SlicerDMRI for it to work. I think perhaps this is why we removed the dependency of SlicerDMRI on UKF. Definitely we want to keep installing UKF when SlicerDMRI is installed, so that is the overall goal.

jhlegarreta:
OK, thanks. IMHO, we should not add dependencies that are not strictly required: e.g. if, for some reason, these latter fail or introduce a regression, then the former extension will be affected and no longer build.

Also, even if `SlicerDMRI`'s `CMakeLists.txt` is listing `UKFTractography` as a dependency, only a warning is being thrown:
https://github.com/jhlegarreta/SlicerDMRI/actions/runs/5719336713/job/15497018205#step:7:168

And building it without is apparently passing:
https://github.com/jhlegarreta/SlicerDMRI/actions/runs/5719336713

Until we come up with some "suggested extension", "dependency of convenience not necessity" as Steve says, or a "meta-extension" mechanism, I would remove it from the dependencies in the `SlicerDMRI`'s `CMakeLists.txt`, and reverting the commit (or making the `HEAD` back to the immediately previous commit, bdb26a14c5f1263f73e5578f581b413d04d00259) would be the most fair thing to do in here. Even more so if there is some reminiscence of a circular dependency.

I should have asked before or marked the PR as WIP. My bad.

pieper:
My feeling is that we should aim to have SlicerDMRI be the one-stop-shop for diffusion imaging the way SlicerRT is for radiotherapy or SlicerIGT is for image guided therapy.  To do that, it would make sense to merge UKFTractography and any other diffusion-related tools into a single extension where the modules work well together.  Of course, this requires that there is a critical mass of developers working together to manage the extension.  It makes more sense to have people working on diffusion to manage their repositories rather than requiring review by the core developers.

jcfr:
> UKFTractography

This is a "soft" dependency and it is not required to build `SlicerDMRI`, considering the `UKFTractography` has been building on all platform, this should not be an issue.

"UKFTractography" can currently be built as a standalone project or as a Slicer extension.

Unless we invest time & resources to simplify it (e.g remove Boost/Eigen/ dependencies[^1], create a dedicated VTK modules [^2], ..), attempting to just "merge" the code into SlicerDMRI may not be that helpful.

**Question**: Is it still relevant to support building "UKFTractography"  as standalone project ? If not, we could start by simplifying the project and facilitate its maintenance.

[^1]: https://github.com/pnlbwh/ukftractography/tree/master/SuperBuild

[^2]: Revisit https://github.com/pnlbwh/ukftractography/tree/master/CompressedSensing, https://github.com/pnlbwh/ukftractography/tree/master/vtkFilter and https://github.com/pnlbwh/ukftractography/tree/master/vtk2mask to create a dedicated VTK modules (potentially also available on PyPI?)
[^2]:

ihnorton:
> @ihnorton please check this. You removed the dependency from the s4ext file back in 2017 - the change was not commented but maybe you still remember.

This was either because of the (lack of) windows build at the time, or a build error that was blocking SliderDMRI from building/releasing. (👋 all)

jhlegarreta:
Thanks for chiming in @ihnorton.

> This was either because of the (lack of) windows build at the time, or a build error that was blocking SliderDMRI from building/releasing

Exactly. I think this is a strong enough argument to advise against adding strictly unnecessary dependencies. Even if these may be building fine for some time, we cannot tell and guarantee, despite having CIs, that this will be the case all the time. So I would reset the `HEAD` to the immediately previous commit.

@jcfr as for https://github.com/Slicer/ExtensionsIndex/pull/1963#issuecomment-1659449872, IMHO there are a number of PRs in the `UKFTractography` repository that first need to be solved before the tool is improved further.

jcfr:
> > **Question:** Is it still relevant to support building "UKFTractography" as standalone project ? If not, we could start by simplifying the project and facilitate its maintenance.

> UKF is frequently used outside of Slicer

@ljod  Thanks for the feedback, that answers the questions.

PR #1971, opened 2023-08-22
BUG: mpReview - update list of extensions to install

deepakri201:
I added the QuantitativeReporting and DICOMwebBrowser extensions to the list of extensions that should be installed when the mpReview extension is installed.


deepakri201:
> 



> @deepakri201 it does not look right that you have 2 .s4ext files. Did you add the new one by mistake?

I believe that @lassoan renamed the s4ext file from SlicermpReview.s4ext to mpReview.s4ext, so there is only one now.

jcfr:
The message seemed to be correct and accurately describe that the description file and the repository name were inconsistent.

@deepakri201 It looks like that instead of updating the existing file [mpReview.s4ext](https://github.com/Slicer/ExtensionsIndex/blob/main/mpReview.s4ext), you are trying to add a new one with the name `SlicermpReview.s4ext`.

For reference, the name was indeed fixed by @lassoan a few weeks ago in https://github.com/Slicer/ExtensionsIndex/commit/290358f84cfc7f015f2283907910a81929416629

I am now closing this pull request.

@deepakri201 To add QuantitativeReporting and DICOMwebBrowserCould as dependency, could you instead edit the existing file [mpReview.s4ext](https://github.com/Slicer/ExtensionsIndex/blob/main/mpReview.s4ext) ?

PR #1972, opened 2023-08-22
BUG: Update mpReview.s4ext

deepakri201:
Added QuantitativeReporting and DICOMwebBrowser as dependencies to mpReview


jcfr:
Thanks 🙏

PR #1984, opened 2023-11-18
Github actions workflow to notify extension managers when there are errors or warnings

vkt1414:


lassoan:
This is really good! Ability to check extension builds and notify developers could be really useful. CDash can send emails if maintainers sign up on [CDash](https://slicer.cdash.org/), but submitting an issue is more visible.
- Is the issue going to be submitted into each extension's issue tracker? (it should be)
- As a start, we should only report regressions (any new test failures)
- We could add a "maintainer" field to the extension description json file (maybe we could require it for tier 3 and above)

PR #1988, opened 2023-12-11
Update TCIABrowser.s4ext

kirbyju:
Fixed links to home page, icon, and screenshot in TCIABrowser.s4ext.  I'm hoping this will make it so the icon actually appears correctly in Slicer after installation as that was not working previously.


pieper (approved):
LGTM 👍

PR #1989, opened 2023-12-11
Request Flywheel Case Iterator for Stable Release

joshicola:
Requesting FlywheelCase Iterator be included in the extension builds for the Slicer 5.4 Stable Release.


lassoan:
Note that extensions are no longer updated for Slicer-5.4. The current stable release is Slicer-5.6.

PR #2021, opened 2024-03-25
Rename Conda to SlicerConda for consistency.

Jeanneclre:
@allemangD 
I renamed the dependencies "Conda", "SlicerConda" because it wasn't found.
In the version 5.7 of Slicer, the extension is named Slicerconda but in the Index, it's called Conda.s4ext. Should we rename the s4ext? 
Thank you :)


allemangD:
Sorry for the commit duplication in 61a5ff9 / 3575948 - moved the file and forgot to add the new one to tracking, and GitHub won't allow me to squash those commits together on the patch branch. When I merge the PR I will squash-and-merge instead.

I agree that renaming the extension to SlicerConda for consistency makes sense... I'm a bit confused that the extension was missing, the dependency name and the `s4ext` file must match. If the file is `Conda.s4ext` then the dependency must be `Conda` - you can see in the CI checks for 5e1948e (and my broken commit 61a5ff9) that the expected SlicerConda.s4ext is missing.

My best guess is that the issue came because the extension project name `SlicerConda` didn't match... I'll keep an eye on CI in the morning to be sure making things consistent fixes the issue.

PR #2023, opened 2024-03-27
ENH: Add SlicerNNUNet extension

Thibault-Pelletier:
# New extension

- [x] LICENSE.txt is present in the repository root. MIT  (https://choosealicense.com/licenses/mit/) or Apache (https://choosealicense.com/licenses/apache-2.0/) license is recommended. If source code license is more restrictive for users than MIT, BSD, Apache, or 3D Slicer license then the name of the used license must be mentioned in the extension description.
  - [x] Extension name
  - [x] License: We suggest you use a permissive license that includes patent and contribution clauses.  This will help protect developers and ensure the code remains freely available.  We suggest you use the [Slicer License](https://github.com/Slicer/Slicer/blob/main/License.txt) or the [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0). Always mention in your README file the license you have chosen.  If you choose a different license, explain why to the extension maintainers. Depending on the license we may not be able to host your work. Read [here](https://opensource.guide/legal/#which-open-source-license-is-appropriate-for-my-project) to learn more about licenses.
- Hide unused features in the repository to reduce noise/irrelevant information:
  - [x] Click `Settings` and in repository settings uncheck `Wiki`, `Projects`, and `Discussions` (if they are currently not used)

### DIFF
--- a/NNUNet.s4ext
+++ b/NNUNet.s4ext
@@ -1,0 +1,44 @@
+#
+# First token of each non-comment line is the keyword and the rest of the line
+# (including spaces) is the value.
+# - the value can be blank
+#
+
+# This is source code manager (i.e. svn)
+scm git
+scmurl https://github.com/KitwareMedical/SlicerNNUnet
+scmrevision main
+
+# list dependencies
+# - These should be names of other modules that have .s4ext files
+# - The dependencies will be built first
+depends PyTorch SlicerPythonTestRunner
+
+# Inner build directory (default is ".")
+build_subdirectory .
+
+# homepage
+homepage https://github.com/KitwareMedical/SlicerNNUnet
+
+# Firstname1 Lastname1 ([SubOrg1, ]Org1), Firstname2 Lastname2 ([SubOrg2, ]Org2)
+# For example: Jane Roe (Superware), John Doe (Lab1, Nowhere), Joe Bloggs (Noware)
+contributors Thibault Pelletier (KITWARE SAS)
+
+# Match category in the xml description of the module (where it shows up in Modules menu)
+category Segmentation
+
+# url to icon (png, size 128x128 pixels)
+iconurl https://github.com/KitwareMedical/SlicerNNUnet/raw/main/SlicerNNUnet/Resources/Icons/SlicerNNUnet.png
+
+# Give people an idea what to expect from this code
+#  - Is it just a test or something you stand behind?
+status stable
+
+# One line stating what the module does
+description 3D Slicer nnUNet integration to streamline usage for nnUNet based AI extensions.
+
+# Space separated list of urls
+screenshoturls https://github.com/KitwareMedical/SlicerNNUnet/raw/main/Screenshots/1.png
+
+# 0 or 1: Define if the extension should be enabled after its installation.
+enabled 1


lassoan:
Thank you, this looks great.

I'm just wondering if the extension name should be `SlicerNNUnet` or `NNUnet`.

In "bridge" extensions, which make available an external tool or library in Slicer used to have the `Slicer` prefix in their name, such as SlicerANTs, SlicerElastix, SlicerFreeSurfer, SlicerJupyter, SlicerOpenAnatomy, SlicerOpenIGTLink, SlicerVMTK, etc. However, this resulted in 25% of the extension names starting with Slicer. Having the `Slicer` prefix for in the extensions catalog is completely redundant and it makes harder to find an extension.

In recent years we tend to move away prefixing everything with Slicer, even in "bridge" extensions. For example, MONAILabel, MONAIAuto3DSeg, MHubRunner, MEMOS, IDCBrowser, NvidiaAIAssistedAnnotation, SPHARM-PDM, TCIABrowser, etc. and it has not resulted in any issues and it made things simpler or shorter.

Overall, I think it would be better to change the extension name to `NNUnet` (by changing the `PROJECT` in the CMakeLists.txt file, and accordingly, update the s4ext filename). Everything else would remain unchanged (repository name SlicerNNUnet, module name SlicerNNUnet, module title nnUNet).

We can discuss any further improvements to the extension that ensures that nnUNet is properly installed and Slicer is not corrupted (e.g., you may need to use [selective pip install](https://github.com/lassoan/SlicerTotalSegmentator/blob/7aa1ca252261a2b6d026cf58d662286aac904fa3/TotalSegmentator/TotalSegmentator.py#L594C9-L594C28) to make sure SimpeITK is not replaced, etc.), but those can be addressed after the extension is added to the Extensions Index.

Thibault-Pelletier:
Hi @lassoan ,

Thanks for the review! I just changed the s4ext to `NNUNet.s4ext` and changed the Project name to simply `NNUNet` in [this commit](https://github.com/KitwareMedical/SlicerNNUnet/commit/0d92ad630834906869e367421f1be6d5de2fe0e5).

Feel free to let me know if you would like me to rename things.

As per our previous discussions on your review, I was not sure if keeping only `NNUNet` and `NNUNetLib` for the module name / library would be confusing for people wanting to use the `nnunet` library instead of the Slicer bridge. I have no strong opinion on that so just let me know what would look best to you.

For the pip install selective, I adapted and improved the logic which was in the Total Segmentator extension 
 [(see here)](https://github.com/KitwareMedical/SlicerNNUnet/blob/main/SlicerNNUnet/SlicerNNUNetLib/InstallLogic.py#L287)
 
The main changes are : 
- I deferred the PyTorch install to make sure the torch required version matches the one present in the nnunetv2 metadata
- I improved the pip install selective to be recursive. Otherwise, sub dependency may install the dependencies which where marked as excluded
- I changed the requirement formatting to use `packaging` module to properly take care of specifiers and install requirements for markers matching platform specific env.

I think that the pip install selective would probably make sense in the `slicer.util` file.

pieper:
My vote would be to call the Slicer extension nnUNet and keep the SlicerNNUNet in python to emphasize that this is the layer where the two systems meet (the code depends on both).  I agree that calling it just nnunet would be confusing.

lassoan:
We all agree thenand this is how it is implemented now.

lassoan:
I've had another look at the extension. You've done a very nice job, there are just a few things to address before this can be approved:
- [ ] It appears that implementation of `pipInstallSelective` is incomplete. It is not enough to skip installation of some dependencies, as dependencies would just creep in or some pip operations would fail. You must actually modify the metadata file of the installed Python package to remove the dependency. `Requires-Dist: ...` lines corresponding to skipped dependencies need to be wiped  (e.g., from `Slicer 5.6.1\lib\Python\Lib\site-packages\nnunetv2-2.3.dist-info\METADATA`). In the past few months, dependencies were mostly OK (except a couple of days) without any package skipping in the Slicer Preview Release, so you may not easily see that selective install does not actually work now. But selective install may become necessary again whenever Python packages in Slicer are updated or nnunet dependencies are changed. If the extension works right now then this does not have to be fixed now, but you need to be aware that most likely the selective install does not actually work well now.
- [ ] according to the template, list of dependent extensions must be `# Specified as a space separated string`. You used comma instead, which might cause problems.
- [ ] I could not test the extension by following the tutorial due to this instruction: "to set the path to the Model to use" because there were no download link to a model that I could have used for testing. Please add a download link to an example model (and if the model does not work on any of the Slicer sample data sets then a sample image as well). We usually store models as a Github release asset, see for example [here](https://github.com/lassoan/SlicerMONAIAuto3DSeg/releases/tag/Models).
- [ ] `Disable TTA` checkbox: Slicer style is to put labels to the left and controls to the right, even for checkboxes. Maybe this decision was not optimal (I don't really like it), but we should follow this style for consistency.
- [ ] Tooltips are good, but I feel a few GUI labels are unnecessarily terse: `NPP` -> `Post-processing pocesses (NPP)`, `Disable TTA` -> `Disable test-time augmentation (TTA)`, same for `NPS`

Selective install could be indeed distributed in Slicer core, but it is quite a hack, it had to be adjusted several times over the past two years when TotalSegmentator suddenly stopped working for some users, and it has not been necessary for other important Python packages (among all the Python packages used by all extensions, only nnunet is important enough and yet has too many and too unstable dependencies so that we had to implement this hack). So, for now it is probably better to be in Slicer extension.

Thibault-Pelletier:
@lassoan thanks for the feedbacks!

I made the changes you listed (including adding a link to the DentalSegmentator weights : https://github.com/gaudot/SlicerDentalSegmentator/releases)

lassoan (approved):
I've tried the extension on Windows and unfortunately it did not work. But it is close enough to be working that we can add it to the extensions index now and I'll submit issues to your repository with details about what needs to be fixed before you start advertise this extension.

Thibault-Pelletier:
Hi @lassoan, 
Thanks for the merge! I saw the issues you opened on the SlicerNNUNet repo. I will address them ASAP (early this week) so that the module is robust enough for the community!

PR #2086, opened 2024-09-07
Remove OpenCV extension until such time as it is needed again

adamrankin:


jamesobutler (commented):
@adamrankin You’ll want to move the file to the Archive directory in this repo and update the Archive README with the date and reasoning for removal (link to discourse post about recent failing build).

PR #2088, opened 2024-09-25
Rename SlicerEditor to ScriptEditor

lassoan:


jcfr:
To reduce confusion, the `SlicerEditor` extension have been manually removed from the `5.6.2` release on `https://slicer-packages.kitware.com` so that it doesn't show up in the extensions manager.

PR #2101, opened 2024-11-04
ENH: Add new dependencies to SlicerMorph.json

smrolfe:
Adds Sandbox and SurfaceMarkup as dependencies to the SlicerMorph extension.

### DIFF
--- a/SlicerMorph.json
+++ b/SlicerMorph.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
-  "build_dependencies": ["SegmentEditorExtraEffects"],
+  "build_dependencies": ["SegmentEditorExtraEffects", "Sandbox", "SurfaceMarkup"],
   "build_subdirectory": ".",
   "category": "SlicerMorph",
   "scm_revision": "master",


SlicerMorph.json
```diff
@@ -1,6 +1,10 @@
 {
   "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
-  "build_dependencies": ["SegmentEditorExtraEffects"],
+  "build_dependencies": [
```
jamesobutler, line 3: https://slicer.readthedocs.io/en/latest/developer_guide/extensions.html#extension-catalog-entry-file

These don’t appear to be build dependencies but rather runtime dependencies? SurfaceMarkup is already specified to be installed when running the PlaceLandmarkGrid module:
https://github.com/SlicerMorph/SlicerMorph/blob/40407085db40774fd976a0bac4c6c287f49b5576/PlaceLandmarkGrid/PlaceLandmarkGrid.py#L48-L64

jamesobutler: Reviewing the Slicer ExtensionsIndex pull request template it appears this change would put it inconsistent with the extensions listed dependencies in its top level CMakeLists.txt.

https://github.com/Slicer/ExtensionsIndex/blob/7cbd60e4339ec82dc297c21710852a13bf53ccaa/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L28

https://github.com/SlicerMorph/SlicerMorph/blob/40407085db40774fd976a0bac4c6c287f49b5576/CMakeLists.txt#L13

smrolfe: Thanks @jamesobutler, these are runtime dependencies and I agree it would be better to install at runtime.

PR #2115, opened 2024-11-21
ENH: Add SurfaceMarkup as a dependency to SlicerMorph.json

smrolfe:
The SurfaceMarkup extension is necessary for the PlaceLandmarkGrid module in SlicerMorph and is currently installed when this module is opened. Since we are now adding support to interact with the generated grids to other modules, this is no longer the best solution. 

Additionally, when running SlicerMorph in a cloud environment, it works best if dependencies are downloaded and installed at the same time as the extension. Adding SurfaceMarkup as an extension dependency will accomplish this also.

### DIFF
--- a/SlicerMorph.json
+++ b/SlicerMorph.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
-  "build_dependencies": ["SegmentEditorExtraEffects"],
+  "build_dependencies": ["SegmentEditorExtraEffects", "SurfaceMarkup"],
   "build_subdirectory": ".",
   "category": "SlicerMorph",
   "scm_revision": "master",


jamesobutler:
@lassoan @jcfr How should we best indicate between build dependencies and runtime dependencies?

lassoan:
Runtime dependencies are managed by the extension. When the user activates a feature that requires an additional Python package, Slicer extension, etc. then the module can display message asking the user to agree to install the extra package/extension and if the user consents then install it.

jamesobutler:
@lassoan Would this mean that an extension that exclusively has Python scripted loadable modules would likely never have a build dependency since the scripted module is doing no compiling that it needs to link against another extension for?

> Additionally, when running SlicerMorph in a cloud environment, it works best if dependencies are downloaded and installed at the same time as the extension. 

@smrolfe Is it possible for the same action that is instructing the installation of the SlicerMorph extension in the cloud environment to also instruct to install the SurfaceMarkup extension? Wouldn’t this generally be a couple lines of code?

Is there also possibility of a concept like the Visual Studio Installer “Workload”? The selection to install the C++ Desktop Environment workload automatically selects related useful components/extensions although these parts don’t necessarily have dependencies of each other. In the Slicer situation it would be able to select a workload that with a single click installs a group of generally useful extensions. In the case of SlicerMorph workload it may install the SlicerMorph extension along with the Sandbox extension if target users of the workload generally fine both usable in the context of each other.

lassoan:
> Would this mean that an extension that exclusively has Python scripted loadable modules would likely never have a Slicer extensions index build dependency since the scripted module is doing no compiling that it needs to link against another extension for?

We could easily add a `runtime_dependencies` property to the [schema](https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json) but right now nothing would use it. We could also use the `runtime_dependencies` as a hint for recommending installing additional extensions, but that could result in installing unnecessary optional extensions (the user may never use those features that require the additional extensions), so I'm not sure this would have much value. If we implement automated testing of multiple extensions then this property will be useful, as we will install the `runtime_dependencies` extensions before starting testing.

For me, adding SurfaceMarkup as a `build dependency` makes sense, because the extension defines additional MRML node types, so if we want to load a scene that contains these node types then the extension is required. It is not a huge issue if we don't add this extension, because if Slicer notices an unknown node type when loading the scene then it will detect that SurfaceMarkup was installed when the scene was saved and it is not currently installed and display this information to the user. However, overall it seems to me that the advantages of adding SurfaceMarkup as a `build dependency` slightly overweighs the disadvantages.

muratmaga:
> @smrolfe Is it possible for the same script that is instructing the installation of the SlicerMorph extension in the cloud environment to also instruct to install the SurfaceMarkup extension? Wouldn’t this generally be a couple lines of code?

The issue is not just cloud though. Basically without defining Surface Markups extension cannot use that markup type. So, to me that is a hard dependency (whether it is runtime or build time is irrelevant in this context).

muratmaga:
Can we get a decision on this?

lassoan:
It seems that you are OK with introducing an unnecessary build dependency because in some cases it makes things slightly more convenient for users. Since this is your extension and also because right now we don't have any specific support for "runtime dependencies" in Slicer core, I think it makes sense to merge this pull request.

I would still recommend to implement a check in your module that uses Surface Markups to check at runtime if the module is installed and if it is not installed (e.g., because the user choose to not install it despite the Extensions Manager offered it; or uninstalled it later) then offer the user to install it (or at least display a meaningful error message).

PR #2119, opened 2024-11-22
ENH: Add Lesion Simulator to Slicer 5.6

acsenrafilho:
This is only an Extension Index update to:

1. Informs the new branch 5.6, which indicates the most recent Slicer stable at the moment
2. Update the repo documentation website, since the Slicer Wiki will be retired in the near future


lassoan (approved):
Thanks for the update.

Issue #2050, opened 2024-06-05
Remove Auto3DGM from extension catalogue


muratmaga:
## Summary
Per developers, Slicer Auto3Dgm extension (https://github.com/ToothAndClaw/SlicerAuto3dgm) is no longer maintained (see the transcript from Doug Boyer below). Their preferred method is now a conda package. 

There are still people who are trying to use this with Slicer. Often it fails and there are also some known issues (such as a dependency for a proprietary library). I would suggest removing it from the extension catalogue so that we don't have to answer the inquiries about why it is not working. 

```
Thanks Murat
 
I think we would probably lean towards dropping it.  Its not really very useful compared to the conda-wrapped python versions we have developed.
Here is the github for the conda/python installation: https://github.com/ToothAndClaw/Auto3dgm_Python.git

Doug 😊
```

jamesobutler:
https://github.com/Slicer/ExtensionsIndex/blob/6b22b3fd0e00abf4207f1cf024b9b1582e1e4a55/Auto3dgm.json#L1-L8

To remove the extension from the extensions index, the json would get added to the [ARCHIVE](https://github.com/Slicer/ExtensionsIndex/tree/6b22b3fd0e00abf4207f1cf024b9b1582e1e4a55/ARCHIVE) and explaining the archive reasoning in the [ARCHIVE README](https://github.com/Slicer/ExtensionsIndex/tree/6b22b3fd0e00abf4207f1cf024b9b1582e1e4a55/ARCHIVE#readme).

Issue #2041, opened 2024-04-24
Meta issue referencing extension updates


jcfr:
_These pull requests were created in the context of https://github.com/Slicer/ExtensionsIndex/pull/2011_

List of pull requests intended to fix and consolidate extension metadata:

* https://github.com/naterex23/SlicerAblationPlanner/pull/3
* https://github.com/jamesobutler/ModelClip/pull/2
* https://github.com/lassoan/AnomalousFiltersExtension/pull/2
* https://github.com/SlicerIGT/PercutaneousApproachAnalysis/pull/1
* https://github.com/QIICR/PkModeling/pull/11
* https://github.com/SlicerIGT/PortPlacement/pull/3
* https://github.com/netstim/SlicerANTs/pull/9
* https://github.com/jcfr/SlicerToKiwiExporter/pull/9
* https://github.com/SlicerDMRI/SlicerWMA-OLD/pull/12
* https://github.com/NrgXnat/XNATSlicer/pull/3
* https://github.com/Slicer/SlicerAirwaySegmentation/pull/1
* https://github.com/DCBIA-OrthoLab/AnglePlanes-Extension/pull/21
* https://github.com/pzaffino/SlicerArduinoController/pull/3
* https://github.com/Atracsys/SlicerAstmPhantomTest/pull/4
* https://github.com/ToothAndClaw/SlicerAuto3dgm/pull/7
* https://github.com/Auditory-Biophysics-Lab/Slicer-BoneThicknessMapping/pull/7
* https://github.com/jamesobutler/SlicerBrainVolumeRefinement/pull/1
* https://github.com/lassoan/SlicerBreastImplantAnalyzer/pull/1
* https://github.com/ikolesov/CarreraSlice/pull/13
* https://github.com/fedorov/ChangeTrackerPy/pull/20
* https://github.com/acil-bwh/SlicerCIP/pull/43
* https://github.com/lassoan/SlicerCleverSegmentation/pull/1
* https://github.com/DCBIA-OrthoLab/CMFreg/pull/39
* https://github.com/tokjun/CurveMaker/pull/17
* https://github.com/DCBIA-OrthoLab/DatabaseInteractorExtension/pull/12
* https://github.com/SlicerRt/SlicerDebuggingTools/pull/17
* https://github.com/gaudot/SlicerDentalSegmentator/pull/2
* https://github.com/Slicer/SlicerDeveloperToolsForExtensions/pull/30
[...]

Issue #2001, opened 2024-02-02
2024.02.02: SlicerBreast_DCEMRI_FTV extension displays a blocking popup


jcfr:
## Observations

It uses `qt.QFileDialog.getExistingDirectory` directly[^1] from module `setup()` function.

[^1]: https://github.com/rnadkarni2/SlicerBreast_DCEMRI_FTV/blob/f0e9ca577ae950f010d6fdb4037195ddaa14b2e9/DCE_IDandPhaseSelect/DCE_IDandPhaseSelect.py#L281-L282

![image](https://github.com/Slicer/ExtensionsIndex/assets/219043/82d6f5a2-570d-415f-b2fa-7394adeee9f9)

## Suggestions

Add a directory selector (e.g `ctk.ctkPathLineEdit` or `ctk.ctkDirectoryButton`) to the UI allowing to select a directory

| `ctkPathLineEdit` | `ctkDirectoryButton` | 
|--|--|
| ![image](https://github.com/Slicer/ExtensionsIndex/assets/219043/7eb38f67-ac23-4f12-89bc-3abe7412deb5) | ![image](https://github.com/Slicer/ExtensionsIndex/assets/219043/95ecaac7-2c07-4ca7-a2c6-1297b08c37ee) |

## Links
* https://github.com/Slicer/ExtensionsIndex/blob/main/Breast_DCEMRI_FTV.s4ext
* https://github.com/rnadkarni2/SlicerBreast_DCEMRI_FTV

Issue #1991, opened 2023-12-12
2023.12.12: An extension displays a blocking popup related to CUDA installation


jcfr:
![image](https://github.com/Slicer/ExtensionsIndex/assets/219043/d307019d-dfc7-4608-abf1-41803efd8626)

lassoan:
This is so wrong in several aspects:
- slicer.util functions should be used for displaying messages (they don't block execution in testing mode)
- no need to exclude Windows or macOS
- no need for low-level GPU access (should be done via SlicerPyTorch)

We can temporarily remove the extension while the maintainers fix the issues.

jcfr:
I will follow-up once identified the culprit.

jcfr:
Here is the code leading the popup:

https://github.com/DCBIA-OrthoLab/SlicerDentalModelSeg/blob/00816c31c570e41a5d580e291519dbf8b541d95e/CrownSegmentation/CrownSegmentation.py#L216-L226

The extension was originally introduced through this pull request:
* https://github.com/Slicer/ExtensionsIndex/pull/1867

The following comment originally posted by @pieper in https://github.com/Slicer/ExtensionsIndex/issues/1867#issuecomment-1206808269 is still relevant:

> Please in the readme be sure to point out that this code depends on CUDA.  It looks like there's already a fallback to CPU mode so describing the tradeoffs in the readme would be helpful.
> 
> Longer term, you should see if you can rework [the installation code](https://github.com/DCBIA-OrthoLab/SlicerDentalModelSeg/blob/main/CrownSegmentationcli/CrownSegmentationcli.py) to use [SlicerPyTorch](https://github.com/fepegar/SlicerPyTorch) to get the right code installed.
> 
> Also it would help if you provide sample data.  I see some in the repository so those should be referenced in the readme or even better registered with the SampleData module so it's easy for users to test the module.
> 
> In the same vein, it looks like you don't have any tests.

The code was originally contributed by @MathieuLeclercq and seems to be maintained by @GaelleLeroux with some input from @allemangD

jcfr:
@@GaelleLeroux With some guidance from us, would you have the bandwidth to help improve this extension ?

GaelleLeroux:
[...]

PR #2123, opened 2024-11-25
Specify "tier" for each extension

lassoan:
By default, Tier 3 is assigned to all extensions - see list of exceptions (that assigned tier 1 or tier 5 instead) in the commit message.

Associated schema update: https://github.com/Slicer/Slicer/pull/8066

Required for https://github.com/Slicer/Slicer/issues/7474

### DIFF
--- a/Auto3dgm.json
+++ b/Auto3dgm.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "Auto3dgm",
   "scm_revision": "master",
-  "scm_url": "https://github.com/ToothAndClaw/SlicerAuto3dgm"
+  "scm_url": "https://github.com/ToothAndClaw/SlicerAuto3dgm",
+  "tier": 1
 }
--- a/Breast_DCEMRI_FTV.json
+++ b/Breast_DCEMRI_FTV.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "FTV Segmentation",
   "scm_revision": "master",
-  "scm_url": "https://github.com/radiology-research/SlicerBreast_DCEMRI_FTV"
+  "scm_url": "https://github.com/radiology-research/SlicerBreast_DCEMRI_FTV",
+  "tier": 1
 }
--- a/ChangeTracker.json
+++ b/ChangeTracker.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "Wizards",
   "scm_revision": "master",
-  "scm_url": "https://github.com/fedorov/ChangeTrackerPy.git"
+  "scm_url": "https://github.com/fedorov/ChangeTrackerPy.git",
+  "tier": 1
 }
--- a/DCMQI.json
+++ b/DCMQI.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": "dcmqi-build",
   "category": "DICOM",
   "scm_revision": "master",
-  "scm_url": "https://github.com/QIICR/dcmqi.git"
+  "scm_url": "https://github.com/QIICR/dcmqi.git",
+  "tier": 5
 }
--- a/ModelClip.json
+++ b/ModelClip.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "Surface Models",
   "scm_revision": "main",
-  "scm_url": "https://github.com/jamesobutler/ModelClip.git"
+  "scm_url": "https://github.com/jamesobutler/ModelClip.git",
+  "tier": 1
 }
--- a/OrthodonticAnalysis.json
+++ b/OrthodonticAnalysis.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "Orthodontics",
   "scm_revision": "master",
-  "scm_url": "https://github.com/OrthodonticAnalysis/SlicerOrthodonticAnalysis"
+  "scm_url": "https://github.com/OrthodonticAnalysis/SlicerOrthodonticAnalysis",
+  "tier": 1
 }
--- a/ShapeVariationAnalyzer.json
+++ b/ShapeVariationAnalyzer.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": ["ShapePopulationViewer"],
   "build_subdirectory": "inner-build",
   "category": "Quantification",
   "scm_revision": "master",
-  "scm_url": "https://github.com/DCBIA-OrthoLab/ShapeVariationAnalyzer.git"
+  "scm_url": "https://github.com/DCBIA-OrthoLab/ShapeVariationAnalyzer.git",
+  "tier": 3
 }
--- a/SlicerDMRI.json
+++ b/SlicerDMRI.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": ["UKFTractography"],
   "build_subdirectory": "inner-build",
   "category": "Diffusion",
   "scm_revision": "master",
-  "scm_url": "https://github.com/SlicerDMRI/SlicerDMRI.git"
+  "scm_url": "https://github.com/SlicerDMRI/SlicerDMRI.git",
+  "tier": 5
 }
--- a/SlicerDentalModelSeg.json
+++ b/SlicerDentalModelSeg.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "Segmentation",
   "scm_revision": "main",
-  "scm_url": "https://github.com/DCBIA-OrthoLab/SlicerDentalModelSeg.git"
+  "scm_url": "https://github.com/DCBIA-OrthoLab/SlicerDentalModelSeg.git",
+  "tier": 1
 }
--- a/SlicerITKUltrasound.json
+++ b/SlicerITKUltrasound.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": "SlicerITKUltrasound-build",
   "category": "Filtering",
   "scm_revision": "master",
-  "scm_url": "https://github.com/KitwareMedical/SlicerITKUltrasound"
+  "scm_url": "https://github.com/KitwareMedical/SlicerITKUltrasound",
+  "tier": 1
 }
--- a/SlicerLookingGlass.json
+++ b/SlicerLookingGlass.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": "inner-build",
   "category": "Holographic Display",
   "scm_revision": "master",
-  "scm_url": "https://github.com/KitwareMedical/SlicerLookingGlass.git"
+  "scm_url": "https://github.com/KitwareMedical/SlicerLookingGlass.git",
+  "tier": 1
 }
--- a/SlicerRadiomics.json
+++ b/SlicerRadiomics.json
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": "inner-build",
   "category": "Informatics",
   "scm_revision": "master",
-  "scm_url": "https://github.com/AIM-Harvard/SlicerRadiomics.git"
+  "scm_url": "https://github.com/AIM-Harvard/SlicerRadiomics.git",
+  "tier": 1
 }
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,27 +6,56 @@ Thank you for contributing to 3D Slicer!
 
 # New extension
 
-<!-- To make sure users can find your extension, understand what it is intended for and how to use it, please complete the checklist below. You do not need to complete all the item by the time you submit the pull request, but most likely the changes will only be merged if all the tasks are done. See more information about the submission process here: https://slicer.readthedocs.io/en/latest/developer_guide/extensions.html -->
+<!-- To make sure users can find your extension, understand what it is intended for and how to use it, please complete the checklist below. You do not need to complete all the item by the time you submit the pull request, but most likely the changes will only be merged if all the tasks are done. "Tier" of your extension will be determined based on the set of requirements you fulfill. See more information about the submission process here: https://slicer.readthedocs.io/en/latest/developer_guide/extensions.html.
+ -->
+
+
+## Tier 1
+
+Any extension that is listed in the Extensions Catalog must fulfill these requirements.
 
 - [ ] Extension has a reasonable name (not too general, not too narrow, suggests what the extension is for)
-- [ ] Repository name is Slicer+ExtensionName
+- [ ] Repository name is Slicer+ExtensionName (except if the repository that hosts the extension can be also used without Slicer)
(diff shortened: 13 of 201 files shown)


DCMQI.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json##",
```
pieper, line 2: Meant to be two `#` characters here?

SlicerLookingGlass.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": "inner-build",
   "category": "Holographic Display",
   "scm_revision": "master",
-  "scm_url": "https://github.com/KitwareMedical/SlicerLookingGlass.git"
+  "scm_url": "https://github.com/KitwareMedical/SlicerLookingGlass.git",
+  "tier": 5
```
pieper, line 8: Last I tried this was not working and my bug report didn't get a reply.  I think this is a 3 or even a 1 until it's confirmed that people are committed to maintaining it.

https://discourse.slicer.org/t/slicerlookingglass-plans/33723/4

pieper (approved):
Thanks for moving ahead with this.  It should help us clear the RP backlog.

There are several json schemas that end with `##` and I'm not sure if that's going to hurt anything, otherwise I think we should go ahead and make any adjustments later.

ModelClip.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "Surface Models",
   "scm_revision": "main",
-  "scm_url": "https://github.com/jamesobutler/ModelClip.git"
+  "scm_url": "https://github.com/jamesobutler/ModelClip.git",
+  "tier": 3
```
jamesobutler, line 8: I would consider this a 1. The fork is being used because the upstream did not respond to a PR with a minor update to fix an extension icon issue.

SlicerRadiomics.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": "inner-build",
   "category": "Informatics",
   "scm_revision": "master",
-  "scm_url": "https://github.com/AIM-Harvard/SlicerRadiomics.git"
+  "scm_url": "https://github.com/AIM-Harvard/SlicerRadiomics.git",
+  "tier": 3
```
jamesobutler, line 8: This one should probably also be considered up for a possible 1 as it has been unavailable for macOS for 1.5 years and users have noticed and asked about it.
https://github.com/AIM-Harvard/SlicerRadiomics/issues/77
https://github.com/AIM-Harvard/SlicerRadiomics/issues/82

SlicerITKUltrasound.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": "SlicerITKUltrasound-build",
   "category": "Filtering",
   "scm_revision": "master",
-  "scm_url": "https://github.com/KitwareMedical/SlicerITKUltrasound"
+  "scm_url": "https://github.com/KitwareMedical/SlicerITKUltrasound",
+  "tier": 3
```
jamesobutler, line 8: This one has long been failing to build across all platforms. If build history is taken into account for tier, this is probably also down to a 1.

@lassoan Other currently failing to build extensions (with a history of failing) should be considered a 1? They don't show up in the extensions manager for users, but even if they did would they have the reputation suddenly of a level 3?

SlicerRadiomics.json
pieper: A 1 might be fair.  It's been unfunded for a very long time and we haven't seen anyone with enough interest to step in an pick it up.

SlicerITKUltrasound.json
lassoan: Yes, long history of failing justifies tier 1

SlicerRadiomics.json
lassoan: It is supposed to be important, but apparently it does not appear to be properly supported. Need to discuss if someone could step up or we acknowledge that this is not avaialable/not important.

SlicerITKUltrasound.json
lassoan: I've set tier to 1

ModelClip.json
lassoan: I've set tier to 1

DCMQI.json
lassoan: fixed

SlicerRadiomics.json
fedorov: I think it is fair to assign any category as you see fit, as long as it is not removed.

Apparently it indeed is important (or at least interesting enough to push the Install button) for quite a few users, based on the download stats.

image

pieper: yes, and since it's only broken on the mac it's not a completely useless extension.

ShapeVariationAnalyzer.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": ["ShapePopulationViewer"],
   "build_subdirectory": "inner-build",
   "category": "Quantification",
   "scm_revision": "master",
-  "scm_url": "https://github.com/DCBIA-OrthoLab/ShapeVariationAnalyzer.git"
+  "scm_url": "https://github.com/DCBIA-OrthoLab/ShapeVariationAnalyzer.git",
+  "tier": 3
```
jamesobutler, line 8: @lassoan Note that this extension was the one that originally brought up the idea of tiers. See https://github.com/Slicer/ExtensionsIndex/issues/1991. Therefore you may actually consider it a 1?

SlicerLookingGlass.json
jcfr: > Let's hear fom @jcfr about the tier for SlicerLookingGlass. I assumed that Kitware is maintaining it, but maybe LookingGlass did not provide funding for the maintenance.

We are interested in maintaining it. One of the challenges is that LookingGlass is developing a new API to support their "Spatial" display[^1], which will likely require a non-trivial amount of work.

Either way, fixing the current extension to support the current API will require updating the VTK version used in Slicer. This update is scheduled for Q1 of 2025. Therefore, for now, setting the tier to 1 is sensible. We will update it once the extension is effectively fixed.

cc: @psavery


[^1]: https://lookingglassfactory.com/16-spatial-oled

jcfr: When we will resume working on the extension, we will then move the file `SlicerLookingGlass.json` out of the `ARCHIVES` while keeping the tier as `1`, once the dashboard will be green, we will then suggest to bump the tier from `1` to `5`.

lassoan: I'll set to tier 1

SlicerDMRI.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": ["UKFTractography"],
   "build_subdirectory": "inner-build",
   "category": "Diffusion",
   "scm_revision": "master",
-  "scm_url": "https://github.com/SlicerDMRI/SlicerDMRI.git"
+  "scm_url": "https://github.com/SlicerDMRI/SlicerDMRI.git",
+  "tier": 5
```
jamesobutler, line 8: As discussed at Slicer hangout, tier 5 extensions should be building appropriately for a new Slicer stable release. Otherwise may drop to tier 3 or 1 if not actively supporting the maintenance.

Currently SlicerDMRI is failing on all platforms for SlicerPreview:
https://slicer.cdash.org/index.php?project=SlicerPreview&date=2024-11-26&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=SlicerDMRI

fedorov:
For the sake of completeness, can someone please link to the documentation that defines tiers and requirements?

lassoan:
> For the sake of completeness, can someone please link to the documentation that defines tiers and requirements?

The requirements are in this pull request, too: https://github.com/Slicer/ExtensionsIndex/pull/2123/files#diff-18813c86948efc57e661623d7ba48ff94325c9b5421ec9177f724922dd553a35

It is mostly about splitting up the existing extension submission checklist, requiring less for tier 1 and requiring all for tier 5 extensions.

SlicerDMRI.json
pieper: Thanks for pointing this out @jamesobutler - This appears to be related to the new dcmtk and we should fix SlicerDMRI accordingly, exactly because it is tier 5.

ChangeTracker.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
```
fedorov, line 2: I suggest this extension is removed. I would be very surprised if it works at all, and I am definitely not supporting it.

Breast_DCEMRI_FTV.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "FTV Segmentation",
   "scm_revision": "master",
-  "scm_url": "https://github.com/radiology-research/SlicerBreast_DCEMRI_FTV"
+  "scm_url": "https://github.com/radiology-research/SlicerBreast_DCEMRI_FTV",
+  "tier": 3
```
jamesobutler, line 8: Based on https://github.com/Slicer/ExtensionsIndex/issues/2001 and corresponding written up issue at https://github.com/radiology-research/SlicerBreast_DCEMRI_FTV/issues/7 without response, this should probably be under consideration for tier 1.

ChangeTracker.json
lassoan: Somebody used it recently and it still worked. I've set it to tier 1 due to no support.

SlicerDMRI.json
lassoan: We'll try to keep at 5 and demote if we cannot fix it for the new Slicer Stable Release.

SlicerRadiomics.json
lassoan: I'll set it to tier 1 because I don't think we can fix the build error anytime soon.

lassoan:
Thanks for all the feedback. I've answered all the comments and checks are passing, so if I don't get any more comments then I'll merge this in about 2 hours.

SlicerDMRI.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": ["UKFTractography"],
```
jamesobutler, line 3: @lassoan UKFTractography itself is listed as Tier 3, but it is a build dependency for Slicer DMRI which is tier 5. If an extension is listed as tier 5 should all of its dependencies be at tier 5 as well?c

In general if an extension is a tier X, then shouldn't all of its dependencies be at the same tier or higher?

ShapeVariationAnalyzer.json
jamesobutler: ah yes. I read the screenshot thinking the mentioned extension behind it was the one. A comment further down explains that the pop-up of #1991 is from SlicerDentalModelSeg

https://github.com/DCBIA-OrthoLab/SlicerDentalModelSeg/blob/00816c31c570e41a5d580e291519dbf8b541d95e/CrownSegmentation/CrownSegmentation.py#L216-L226

ChangeTracker.json
fedorov: Wow, this is incredible! If this rather involved extension is still working, with close to zero maintenance over the past 5+ years (at least!) it would speak tons about the stability of Slicer platform.

But of course, in this case tier 1 is ok with me!

SlicerDMRI.json
pieper: +1 for UKF as tier 5, and yes, in general to the transitive property of tiers (that their dependencies should have equal or higher tiers).

But since we are just introducing tiers I don't think UKF should be a blocker for the release, but it should be fixed asap.

SlicerDentalModelSeg.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "Segmentation",
   "scm_revision": "main",
-  "scm_url": "https://github.com/DCBIA-OrthoLab/SlicerDentalModelSeg.git"
+  "scm_url": "https://github.com/DCBIA-OrthoLab/SlicerDentalModelSeg.git",
+  "tier": 3
```
jamesobutler, line 8: #1991 was created because of this SlicerDentalModelSeg extension. It was brought up again by @jcfr in https://github.com/Slicer/ExtensionsIndex/issues/1991#issuecomment-2182821199.

https://github.com/DCBIA-OrthoLab/SlicerDentalModelSeg/blob/9bfce1f878abd5ff60a81d275b7292babc697fbe/CrownSegmentation/CrownSegmentation.py#L276-L278

It reports itself as working for Linux only which would probably make it a tier 1 as well since the platform has the fewest number of users compared to macOS and Windows?

SlicerDMRI.json
jamesobutler: Regarding dependencies -- probably good to add to the Tier documentation something about all dependencies must meet the same tier or higher when grading the tier of the specific extension.

SlicerLookingGlass.json
jamesobutler: @lassoan Reminder to update your commit message body which currently lists `SlicerLookingGlass` as Tier 5.

SlicerDMRI.json
jamesobutler: Focusing on the currently specified Tier 5 extensions, the following have less than Tier 5 build dependencies.

- SlicerMorph has a Tier 3 dependency of SurfaceMarkup
- SlicerNNUNet has a Tier 3 dependency of SlicerPythonTestRunner
- SlicerVMTK has a Tier 3 dependency of ExtraMarkups
- QuantitativeReporting has multiple Tier 3 dependencies in SlicerDevelopmentToolbox and PETDICOMExtension
- SlicerNeuro has multiple Tier 3 dependencies in UKFTractography, SwissSkullStripper, HDBrainExtension

SlicerLookingGlass.json
lassoan: Thanks good catch, I'll remove the name from tier 5 (it is already listed under tier 1).

SlicerDMRI.json
lassoan: Tiers don't need to propagate to dependencies. Even though one extension may use some modules from another extension, the extensions are developed, supported, tested, documented separately. Tier is determined based on the merit of each extension, not based what dependencies it uses. If a tier 5 extension depends on a tier 1 extension then its the responsibility of tier 5 extension developers to make sure that the extension fulfills all tier 5 requirements, so for example if a tier 1 dependency fails to build then the tier 5 maintainers may need to jump in and fix some issues or remove the dependency.

SlicerDentalModelSeg.json
lassoan: I've moved UKFTractography to tier 5

SlicerDMRI.json
lassoan: I've moved UKFTractography to tier 5.

jamesobutler: I think it would be important for the top tier (Tier 5) extensions to have a solid foundation of dependencies, so if it is using a Tier 1 extension it should probably be at least graded at Tier 3. It would be poor form for a top critical to the community extension to be using a poor Tier 1 extension. To increase its score to Tier 5, those developers should improve the quality of the dependency to reach Tier 5.

SlicerDentalModelSeg.json
lassoan: I see, thanks, I'll set tier to 1.

SlicerDMRI.json
lassoan: > It would be poor form for a top critical to the community extension to be using a poor Tier 1 extension.

I agree, but I would trust the judgement of the extension maintainers to choose their dependencies carefully. Developers are already responsible for choosing and maintaining their other dependencies (C++ libraries, Python packages, etc.) I don't think we could or should regulate extension dependencies differently.

jamesobutler: Yeah I just think a Tier 5 extensions with dependencies of other Tier 5 extensions is uniquely better than if an extension was Tier 5 with Tier1/3 dependencies. To make it clear that the first case is the best it would mean keeping Tier 5 truly the best of the best and most extensions probably at the Tier 3 level to not water down the Tier 5 category.

lassoan: Yes, it is better if an extension does not have dependencies or they are already at the same or higher tier and we should really work towards that.

However, if SlicerMorph (tier 5) developers want to start using some features from SurfaceMarkup extension (tier 3), then what could you do about it? We cannot prohibit SlicerMorph developers from using SurfaceMarkups, you cannot force them to bring SurfaceMarkup extension up to tier 5 either, and you cannot reduce SlicerMorph tier if they take responsibility of fulfilling all the tier 5 requirements.

jamesobutler: I think SlicerMorph would become a Tier 3 extension (based on Tier 5 detailing all dependent extensions also be Tier 5). I agree you can’t force them to do something, but the grading could be changed. I don’t see any current negatives of being Tier 3 as it will still be visible like Tier 5 extensions. Only the poor Tier 1 extensions will be hidden by default.

lassoan: If SlicerMorph wants to use a tier 1 extension, would you bring down SlicerMorph to tier 1? Even if the dependency is just a runtime dependency, for a specific feature of a single module?

Who and how could enforce these dependency rules? (Slicer maintainers do not know when runtime dependencies in an extension change)

jamesobutler: I think what I’m focused on here is the Slicer build dependencies as specified in the schema and not the runtime dependencies. I have less concern of getting Tier 1 and 3 correct, but Tier 5 should really be the best of the best in terms of code, maintainability and support. SlicerMorph is top tier for responsiveness, but if it uses an extension that doesn’t have great automated tests or doesn’t have write access for a core Slicer dev, then seems reasonable to have it be Tier 3. If it has a build dependency that is Linux only, then at that point it would make sense to bring it all the way down to Tier 1.

Auto3dgm.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "Auto3dgm",
   "scm_revision": "master",
-  "scm_url": "https://github.com/ToothAndClaw/SlicerAuto3dgm"
+  "scm_url": "https://github.com/ToothAndClaw/SlicerAuto3dgm",
+  "tier": 3
```
jamesobutler, line 8: Based on https://github.com/Slicer/ExtensionsIndex/issues/2050, the developers are no longer supporting it, often it fails and an indication to remove it from the Slicer extensions index entirely:
```suggestion
  "tier": 1
```

OrthodonticAnalysis.json
```diff
@@ -1,8 +1,9 @@
 {
-  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.0.json#",
+  "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "Orthodontics",
   "scm_revision": "master",
-  "scm_url": "https://github.com/OrthodonticAnalysis/SlicerOrthodonticAnalysis"
+  "scm_url": "https://github.com/OrthodonticAnalysis/SlicerOrthodonticAnalysis",
+  "tier": 3
```
jamesobutler, line 8: Upon reviewing linked PRs mentioned in https://github.com/Slicer/ExtensionsIndex/issues/2041#issue-2260232701, based on the lack of maintainer response to https://github.com/OrthodonticAnalysis/SlicerOrthodonticAnalysis/pull/4, this could be under consideration for tier 1.

lassoan: > Upon reviewing linked PRs mentioned in [#2041 (comment)](https://github.com/Slicer/ExtensionsIndex/issues/2041#issue-2260232701), based on the lack of maintainer response to [OrthodonticAnalysis/SlicerOrthodonticAnalysis#4](https://github.com/OrthodonticAnalysis/SlicerOrthodonticAnalysis/pull/4), this could be under consideration for tier 1.

OK, I'll set it to 1

jamesobutler:
> Current tier 5 list is aspirational

Ok. Yeah as it relates to the user experience having the tiers represent the state of the extensions that they currently are in rather than what we would hope them to be would be the most beneficial. Ultimately the user will have a poor experience if a Tier 5 is non-functional as they will lose trust in the rankings. Life happens and developers can’t always provide maintenance to an extension due to not having enough time or resources to fund the effort. There will definitely be at times where something we really want to have be available say SlicerDMRI and UKFTractography as Tier 5, but there isn’t the resources associated with a Tier 5 extension to reflect it to users. I think most extension developers would aspire to be Tier 5 even though realistically they are unable to provide that Tier 5 experience to the users consistently.

jamesobutler:
@lassoan It appears that SlicerNeuropacs was not given a tier ranking in this PR. SlicerNeuropacs extension was added to the index prior to this PR being integrated.
https://github.com/Slicer/ExtensionsIndex/blob/bb178bf37659e29586ff768b38bb9aa692998f12/SlicerNeuropacs.json#L1-L8

PR #2124, opened 2024-12-02
Do not require "good" extension name for Tier 1

lassoan:
It is often hard to judge if an extension's name is appropriate, when currently the scope of the extension is narrow but developers plan to broaden it in the future.

For example, `SegmentHumanBody` name is extremely broad, but it could be justified for an extension that proves to be universally useful for segmenting anything in the human body (such as TotalSegmentator or MONAIAuto3DSeg).

This commit would allow accepting an extension with a very ambitious name as Tier 1. Since Tier 1 extensions will not be shown by default, an extension with a too ambitious name would not cause disappointments.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -14,7 +14,6 @@ Thank you for contributing to 3D Slicer!
 
 Any extension that is listed in the Extensions Catalog must fulfill these requirements.
 
-- [ ] Extension has a reasonable name (not too general, not too narrow, suggests what the extension is for)
 - [ ] Repository name is Slicer+ExtensionName (except if the repository that hosts the extension can be also used without Slicer)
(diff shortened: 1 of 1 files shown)


pieper (approved):
Yes, makes sense to me.

pieper:
Not sure what's doing on with the link error...

jamesobutler:
The 

> FailedFileLoadError: Failed to parse 

appears to not have been fully resolved by https://github.com/Slicer/ExtensionsIndex/commit/7b3823f78f76c71fc3139da8f0fc92f4253ff72f based on it happening again in a run for this PR(https://github.com/Slicer/ExtensionsIndex/actions/runs/12128642830/job/33815438357?pr=2124). 

Re-running the failed job suddenly resulted in it passing. See https://github.com/Slicer/ExtensionsIndex/actions/runs/12128642830/job/33819733020.

jamesobutler (approved):

PR #2143, opened 2025-02-15
Fix mismatch of Photogrammetry extensions json and extension name

jamesobutler:
This fixes up accidental mistakes integrated through https://github.com/Slicer/ExtensionsIndex/pull/2139 where the json file was "SlicerPhotogrammetry" while the extension is "Photogrammetry" (per [here](https://github.com/SlicerMorph/SlicerPhotogrammetry/blob/6d65f18f7416e11a4e2e4acb9b44ece4256cf5f9/CMakeLists.txt#L2)).


pieper:
@jamesobutler looks like the same issue here?

https://discourse.slicer.org/t/how-to-install-slicer-dti-alps-extension/41713/2

https://slicer.cdash.org/viewBuildError.php?buildid=3690013

Maybe we can find all cases like this with a script.

PR #2150, opened 2025-03-18
ENH: point to a specified branch anticipating a markups JSON API change.

chir-set:
See:
  https://github.com/chir-set/SlicerExtraMarkups/pull/9

Slicer [PR #8141](https://github.com/Slicer/Slicer/pull/8141) will introduce backwards incompatible changes.

Point the 5.8 stable branch of Slicer to the 'SlicerStable' branch of ExtraMarkups.

### DIFF
--- a/ExtraMarkups.json
+++ b/ExtraMarkups.json
@@ -3,7 +3,7 @@
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "Utilities",
-  "scm_revision": "main",
+  "scm_revision": "5.8",
   "scm_url": "https://github.com/chir-set/SlicerExtraMarkups.git",
   "tier": 3
 }


ExtraMarkups.json
```diff
@@ -3,7 +3,7 @@
   "build_dependencies": [],
   "build_subdirectory": ".",
   "category": "Utilities",
-  "scm_revision": "main",
+  "scm_revision": "SlicerStable",
```
lassoan, line 6: `SlicerStable` is not a suitable branch name as its meaning changes in time. Right now SlicerStable means 5.8 but in 4 months it will mean 5.10. I would recommend to follow the convention that is used in almost all extensions: use the Slicer branch name as branch name in the extension's repository.
```suggestion
  "scm_revision": "5.8",
```

chir-set: Yep, you're right, it's fixed now.

lassoan:
Thank you! Would you mind referring to the extension's name in the commit message? It helps reading/searching the commit history if the extension name is mentioned.

chir-set:
> Thank you! Would you mind referring to the extension's name in the commit message? It helps reading/searching the commit history if the extension name is mentioned.

That's done, just a name, not a link.

chir-set:
A lint check fails with "5.8", which seems a legitimate value. I have no idea if this is a blocking failure.

lassoan:
> A lint check fails with "5.8", which seems a legitimate value. I have no idea if this is a blocking failure.

No problem, sometimes the json schema check fails (probably the server that hosts the schema file is not robust). The commit was perfect and now it has been merged.

PR #2177, opened 2025-05-30
ENH: freeze dcmqi version to v1.4.0

fedorov:
This is needed due to new developments in dcmqi that will require a more recent DCMTK 3.6.9, while Slicer is currently on 3.6.8.


fedorov:
@jcfr @lassoan I don't understand the lint error. Can you let me know what I am doing incorrectly?

lassoan:
The schema validation has trouble downloading the schema. The change looks good, merging.

PR #2193, opened 2025-08-08
BUG: Change Trame.json to SlicerTrame.json

jcfr:
Renames the extension catalog entry file from `Trame.json` to `SlicerTrame.json` for consistency with the extension's advertised[^1] purpose and name. This avoids confusion with the broader *trame* project.

[^1]: https://github.com/KitwareMedical/SlicerTrame?tab=readme-ov-file#slicertrame

This change also addresses the CI packaging failure [reported on CDash](https://slicer.cdash.org/builds/3887961):

```
Package and upload extension
CPack log: /.../Trame-build/inner-build/packageupload_cpack_output.txt
CMake Error at C:/D/P/S-0/Extensions/CMake/SlicerFunctionExtractExtensionDescription.cmake:135 (message):
CUSTOMBUILD : error : EXTENSION_FILE CMake variable points to a inexistent file or [C:\D\P\S-0-E-b\Trame-build\inner-build\packageupload.vcxproj]
    directory: /.../Trame-build/inner-build/SlicerTrame.json
Call Stack (most recent call first):
  C:/D/P/S-0/Extensions/CMake/SlicerExtensionPackageAndUploadTarget.cmake:210 (slicerFunctionExtractExtensionDescriptionFromJson)
```

By aligning the file name with the actual extension name, the build process can correctly locate the JSON descriptor, unblocking packaging.

> [!IMPORTANT]
> Consider reverting [86ef52b](https://github.com/KitwareMedical/SlicerTrame/commit/86ef52bb76deabc0064e27fbd21e297647c10ba0) in the SlicerTrame repository for consistent naming. A force-push may be appropriate.

Related pull requests:
* https://github.com/Slicer/ExtensionsIndex/pull/2174

@Thibault-Pelletier


Thibault-Pelletier:
Thanks for the push!
This looks good to me. I removed the renaming from the SlicerTrame repo and kept the CMake rework to make the different CMake variables follow the EXTENSION_NAME instead of the raw SlicerTrame string.

Thibault-Pelletier (approved):

lassoan:
In general, using "Slicer" as a prefix in a Slicer extension name is redundant and makes it harder to find extensions (as your need to look for Abc and SlicerAbc in the extensions list). But this might be a rare case where it makes sense - if we expect to have various trame applications in the extensions index (SlicerTrame, SimpleSegmenterTrame, CardioViewTrame, ...), or if we intend to provide a new general-purpose trame extension in the future where the user can choose between several applications.

If we anticipate to only offer Slicer as a trame application in the extension manager then it is better to keep the simple "Trame" name. Or maybe it would be even better for users (easier to discove, easier to remember) to choose a name that reflects what the extension is used for (WebUI, RemoteAccess, etc) instead of naming it based on some of the underlying technology that the user has never heard of. It could also make sense to convince the two, such as TrameWebUI to advertise the trame brand while also allow users to guess what the extension could be about.

PR #2195, opened 2025-08-13
Update MorphoDepot.json - pin version

pieper:
Allow testing and fixes to continue on main branch.

Since we are using the Slicer Preview releases in production due to the need for Terminology changes, we should treat this version as stable for now and not get untested changes into the published extension.

This is needed since I pushed a few commits to MorphoDepot main branch before realizing that if they have any issues they could disrupt current users.

### DIFF
--- a/MorphoDepot.json
+++ b/MorphoDepot.json
@@ -5,7 +5,7 @@
   ],
   "build_subdirectory": ".",
   "category": "SlicerMorph",
-  "scm_revision": "main",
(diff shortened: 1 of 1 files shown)


pieper:
It would be great if someone could review and approve, just to stick with the process, but if not I'll merge this later just so the extension stays unchanged for the previews.

jamesobutler (approved):

PR #2207, opened 2025-09-06
Clarified requirements for extension name

lassoan:
Lots of submitted extensions has not very good names. Updated the checklist to help with that.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -14,6 +14,7 @@ Thank you for contributing to 3D Slicer!
 
 Any extension that is listed in the Extensions Catalog must fulfill these requirements.
 
+- [ ] Extension has a reasonable name (not too general, not too narrow, suggests what the extension is for). The extension name should not start with `Slicer`, because it would make it more difficult to find extensions if the name of many started with the same word.
 - [ ] Repository name is Slicer+ExtensionName (except if the repository that hosts the extension can be also used without Slicer)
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -14,6 +14,7 @@ Thank you for contributing to 3D Slicer!
 
 Any extension that is listed in the Extensions Catalog must fulfill these requirements.
 
+- [ ] Extension has a reasonable name (not too general, not too narrow, suggests what the extension is for). The extension name should not start with `Slicer`, because it would make it more difficult to find extensions if the name of many started with the same word.
```
jcfr, line 17: I suggest to tweak like this:

```
... not start with `Slicer` (unless it explicitly provides a bridge between Slicer and a tools or library), because ....
```

(Ps: reviewing from my phone on which i can't do proper suggestions)

lassoan:
The check involves checking the repository structure (no extra folders, files, etc.) so it is somewhat useful if the tests are run. If no modified json files are found then most tests are skipped.

.github/PULL_REQUEST_TEMPLATE.md
lassoan, line 17:
```suggestion
- [ ] Extension has a reasonable name (not too general, not too narrow, suggests what the extension is for). The extension name should not start with `Slicer` (unless it explicitly provides a bridge between Slicer and a tool or library), because it would make it more difficult to find extensions if the name of many started with the same word.
```

PR #2220, opened 2025-09-09
Update BoneReconstructionPlanner.json

mauigna06:
Add Telemetry extension as a dependency on `main` branch


mauigna06:
I did not mean to force-push over your commit Andras, I had wrote the wrong name for the SlicerTelemetry extension so I had to fix it :)

PR #2224, opened 2025-09-14
Removed Telemetry dependency from BoneReconstructionPlanner

lassoan:
Slicer build fails due to Telemetry extension is not in the ExtensionsIndex but required by BoneReconstructionPlanner. Since Slicer-5.8 builds (and therefore extension updates) will stop within in a week it is better to not add the Telemetry extension now (which is still not widely tested and so may not be very mature). It is not a build-time dependency anyway and its installation is optional for the user.

FYI @mauigna06


mauigna06:
Ok, thanks for letting me know

PR #2226, opened 2025-09-15
BUG: Fix extension index build updating LayerDisplayableManager build_subdirectory

jcfr:
This fixes the following error reported when attempting to build the extensions index:

```
-- Configuring extension: LayerDisplayableManager
CMake Error at SlicerBlockUploadExtension.cmake:23 (message):
  Variable EXTENSION_BUILD_SUBDIRECTORY is not defined !
Call Stack (most recent call first):
  SlicerBlockBuildPackageAndUploadExtensions.cmake:187 (include)
  CMakeLists.txt:63 (include)
```

### DIFF
--- a/LayerDisplayableManager.json
+++ b/LayerDisplayableManager.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [],
-  "build_subdirectory": "",
+  "build_subdirectory": ".",
   "category": "Developer Tools",
   "scm_revision": "main",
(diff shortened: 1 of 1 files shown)


jcfr:
Follow-up of:
* https://github.com/Slicer/ExtensionsIndex/pull/2223

Thibault-Pelletier:
Thanks for the fix! 🙏

jamesobutler:
If `build_subdirectory` has a default of ".", then the recommendation to the user should probably be:
- "remove `build_subdirectory` property" if the repo has no "Superbuild" named directory
- "update `build_subdirectory` property" to "inner-build", if the repo has a "Superbuild" named directory and the property is not set to "inner-build" (unspecified or an invalid specification like "").

PR #2331, opened 2026-04-08
Add SlicerConnect Extension

pka420:
# New extension

## Tier 1

- [x] Extension has a reasonable name (not too general, not too narrow, suggests what the extension is for). The extension name should not start with `Slicer` (unless it explicitly provides a bridge between Slicer and a tool or library), because it would make it more difficult to find extensions if the name of many started with the same word.
  - [x] Extension name
- The extension is safe:

### DIFF
--- a/SlicerConnect.json
+++ b/SlicerConnect.json
@@ -1,0 +1,13 @@
+{
+  "name": "SlicerConnect",
+  "category": "Collaboration",
+  "description": "Real-time multi-user collaboration for medical image segmentation.",
+  "scm_url": "https://github.com/pka420/SlicerConnect.git",
+  "scm_revision": "main",
+  "scm_type": "git",
+  "contributors": "Piyush Khurana (Delhi University)",
+  "homepage": "https://github.com/pka420/SlicerConnect",
+  "icon_url": "https://raw.githubusercontent.com/pka420/SlicerConnect/main/SlicerConnect.png",
+  "build_dependencies": [],
+  "enabled": true
+}


pieper:
This sounds pretty cool to me 👍

pka420:
> This sounds pretty cool to me 👍

Thanks! It is buggy though and in experimental mode,
Do I need to do anything for this to be merged?

pka420:
@pieper Please let me know if I need anything else to do here.

pieper:
This sounds like a cool project.  There are a couple checks not passing (see below).

Also it looks like there are some other things to address (the text below is Claude Code talking, not me).


Quick take: the scope and structure are reasonable for a Slicer extension (three ScriptedLoadableModules, proper CMakeLists, UI files, logic/widget separation), and the delta-sync idea is sound. But the code has several shipping blockers and a handful of Slicer-idiom problems you'd want fixed before calling this usable — and more before submitting to the ExtensionsIndex.

Shipping blockers (will crash or not build):

[SlicerConnect.py:67](https://github.com/pka420/SlicerConnect/blob/main/SlicerConnect/SlicerConnect.py#L67) — stray bareword m inside _try_refresh_token; every 401 refresh path raises NameError.
[SlicerConnect.py:488](https://github.com/pka420/SlicerConnect/blob/main/SlicerConnect/SlicerConnect.py#L488) — local_path=Null (should be None); download flow can't execute.
[SlicerConnect.py:148](https://github.com/pka420/SlicerConnect/blob/main/SlicerConnect/SlicerConnect.py#L148) — uses json.dumps but json is never imported.
[SlicerConnectLogin.py:150](https://github.com/pka420/SlicerConnect/blob/main/SlicerConnectLogin/SlicerConnectLogin.py#L150) — selectModule("CollaborativeSegmentation") — no such module; should be "SlicerConnect".
Top-level [CMakeLists.txt](https://github.com/pka420/SlicerConnect/blob/main/CMakeLists.txt) — EXTENSION_LICENSE_FILE points to LICENSE but repo has LICENSE.txt; ICONURL/SCREENSHOT_URLS are absolute root paths, need full URLs for the index.
Security/privacy — contradicts the README:

[SlicerConnectEditor.py:342](https://github.com/pka420/SlicerConnect/blob/main/SlicerConnectEditor/SlicerConnectEditor.py#L342) uses ws:// (cleartext) while README promises WSS. For PHI claims, that's a real problem.
Auth token appended as a URL query param on the WS handshake — ends up in proxy/server logs. Use a subprotocol or Authorization header.
Token is persisted to slicer.app.settings() (plaintext QSettings).
Slicer-idiom issues:

Main-thread pip_install + synchronous requests calls block the UI.
QTimer-polled non-blocking websocket-client at 200 ms works but QWebSocket would integrate with Qt's event loop without polling.
_applyArrayToSegmentation rebuilds every segment on each delta, largely defeating the delta optimization, and the per-label flatten drops Slicer's native overlapping-segment support.
SlicerConnectEditor category "None" (literal string) — it'll appear under a category named "None".
slicer.util.currentTime() [SlicerConnect.py:573](https://github.com/pka420/SlicerConnect/blob/main/SlicerConnect/SlicerConnect.py#L573) isn't a real API.
File handles opened in files={"file": (..., open(path, "rb"), ...)} are never closed.
Design smell: the _resampleToShape + "shape mismatch" branch is papering over a bug — a bound session shouldn't see shape changes. Worth fixing the root cause instead.

Overall: the architecture is credible, but it reads like an early-stage prototype that hasn't been end-to-end exercised — several of the blockers would surface the first time you clicked a button. Worth a cleanup pass before it's demo-worthy, and real encryption/credential handling before anyone touches patient data with it.

pieper:
BTW, for pure python modules you really don't need to make extensions at all.  You can just put install instructions on your github page and post about the modules on the slicer forum.  As long as you follow the extension format, all people need to do is checkout your code (or download the zip) and drop the folder on Slicer.  Don't even need a restart.  We should probably make this more clear in the dev instructions.

pka420:
> an early-stage prototype that hasn't been end-to-end exercised — several of the blockers would surface the first time you clicked a button. Worth a cleanup pass before it's demo-worthy, and real encryption/credential handling before anyone touches patient data with it.

Thanks! I will work on the fix.
I do want it to publish the extension (even if its pure python) to showcase the project.

lassoan:
Nice work! Real-time multi-user collaboration so far has been mostly available in Slicer-based commercial applications and a few open-source research prototypes, so it is useful to have an up-to-date, maintained open-source solution for this. I would recommend to advertise the open-source server as well (https://github.com/pka420/SlicerConnectBackend) as many people may not want to rely on a proprietary server.

Make sure that all the automatic tests in "Extension Validation" and "Lint" actions are fixed, we cannot merge if those tests are failing.

Consider uploading a more informative screenshot (the current one shows a single empty Slicer application window).

PR #2340, opened 2026-04-17
New Extension SegmentComparison

fbrewster:
# New extension

## Tier 1

- [x] Extension has a reasonable name (not too general, not too narrow, suggests what the extension is for). The extension name should not start with `Slicer` (unless it explicitly provides a bridge between Slicer and a tool or library), because it would make it more difficult to find extensions if the name of many started with the same word.
  - [x] Extension name
- The extension is safe:

## Tier 3

Community-supported extensions.

- [x] Documentation, tutorial, and test data are provided for most modules. A tutorial provides step-by-step description of at least the most typical use case, include a few screenshots. Any sample data sets that is used in tutorials must be registered with the Sample Data module to provide easy access to the user.
- [x] Follows programming and user interface conventions of 3D Slicer (e.g., GUI and logic are separated, usage of popups is minimized, no unnecessary custom GUI styling, etc.)
- [x] The extension can be successfully built and packaged on all supported platforms (Windows, macOS, Linux)
- [x] Maintainers respond to issues and pull request submitted to the extension's repository.
- [x] Maintainers respond to questions directly addressed to him/her via @mention on the [Slicer Forum](https://discourse.slicer.org).
- [x] Permissive license is used for the main functions of the extension (recommended Apache or MIT). The extension can provide additional functionality in optional components that are distributed with non-permissive license, but the user has to explicitly approve those before using them (e.g., a pop-up can be displayed that explains the licensing terms and the user has to acknowledge them to proceed).
- All requirements of tiers < 3.

## Tier 5

Critically important extensions, supported by Slicer core developers. New Slicer Stable Release is released only if all Tier 5 extension packages are successfully created on all supported platforms.

- [ ] Slicer core developers accept the responsibility of fixing any issues caused by Slicer core changes; at least one Slicer core developer (anyone who has commit right to Slicer core) must be granted commit right to the extension's repository.
- [ ] Automated tests for all critical features.
- [ ] Maintainers respond to questions related to the extension on the [Slicer Forum](https://discourse.slicer.org).
- All requirements of tiers < 5.


pieper:
This looks cool, thanks for submitting it 👍 

I didn't look at the code closely, but one thing we do often iterate on is the name of the extension so that users can get a quick idea what it's for.  In this case I think SegmentComparison or something like that would get the idea across better.

fbrewster:
Sure, agreed that's more descriptive! I've updated now

fbrewster:
Hi @sjh26, I've updated the name on this and then had to make some changes to the CMake file (it wasn't happy with a name with spaces despite being in quotes) but should be good to go now. Am I just waiting for the workflow to run or have I done something stupid?

cpinter:
Sorry just seeinhg the conversation about the name. We have a SegmentComparison module in SlicerRT that implements the more typical Dice and Hausdorff segment comparison metrics. This new extension (if I saw correctly) adds two different methods. Its name could suggest that it is THE extension containing segment comparison tools, when in reality it provides additional ones.

I don't have any great suggestions, but based on how we do this with the segment editor effects (basic ones in Segment Editor module and additional ones in SegmentEditorExtraEffects), it could be SegmentComparisonExtra or SegmentComparisonAddon, something like this.

lassoan:
For users (and probably in the long term for the maintainers, too), the simplest would be to have one SegmentComparison extension.

@fbrewster would you consider adding your modules to the existing SegmentComparison extension?

jamesobutler:
@lassoan There is an existing SegmentComparison extension? Or are you thinking of the SegmentationReview extension (https://github.com/zapaishchykova/SegmentationReview)? 

It would seem reasonable to pull the SegmentComparison module out of SlicerRT into its own extension for general use and reuse by various AI workflows since it isn’t RT specific.

lassoan:
Sorry, I did not remember correctly - as Csaba wrote as well, [SegmentComparison](https://github.com/SlicerRt/SlicerRT/tree/master/SegmentComparison) is just a module, not a separate extension.

I agree with @jamesobutler suggestion - What do you all think about creating a `SegmentComparison` extension by moving SegmentComparison module from SlicerRT and add @fbrewster's additional methods? There is a much better chance that users would find segment comparison modules if there was a dedicated extension for them.

cpinter:
> It would seem reasonable to pull the SegmentComparison module out of SlicerRT into its own extension for general use

There is a very practical reason for not having done this yet, although we thought it useful a long time ago. The Dice and Hausdorff metrics are implemented (and well tested, validated) in Plastimatch, which is a really bulky dependency of SlicerRT, also used for a bunch of other things (DICOM-RT support, proton dose computation, gamma dose comparison, etc.).

It has always seemed a huge task leaving Plastimatch behind, but maybe we can start approaching now that we have the powerful AI tools. The most unique feature, which is the proton calculation, is somewhat unstable (crashes more than not when run on larger CTs), and is foreseeable to be replaced by the engine in pyRadPlan (already integrated into SlicerRT). I can collect the list of functions to replace and then re-implement in VTK using Claude for example. We already have tests for the majority of the Plastimatch provided functions, but probably we'll need to add more automated tests before making the change, and we'll need to make sure the new implementations are bullet proof.

What do you think?

fbrewster:
Sorry just catching up on this. Happy with this being rolled into the SegmentComparison module of SlicerRT or with it being pulled into a separate extension. I agree that makes sense conceptually.

The BLD module would be easily converted into a separate one for Hausdorff distance (+/-95%) if we went with surface-based instead of a distance map (uses [SciPy's KDTree](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.KDTree.html)), looks like Plastimatch makes a distance map though.

PR #2345, opened 2026-04-27
ENH: remove dcmqi extension

fedorov:
It should be included in the Slicer core now that https://github.com/Slicer/Slicer/pull/9114 is merged.

QuantitativeReporting dependency on dcmqi was removed via https://github.com/QIICR/QuantitativeReporting/pull/297.

PETLiverUptakeMeasurement moved to archive until it is updated to not depend on the DCMQI extension.


ARCHIVE/PETLiverUptakeMeasurement.json
```diff
@@ -1,7 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/Slicer/Slicer/main/Schemas/slicer-extension-catalog-entry-schema-v1.0.1.json#",
   "build_dependencies": [
-    "DCMQI"
```
jamesobutler, line 4: PETLiverUptakeExtension still has a CMakeLists.txt (as of latest master https://github.com/QIICR/PETLiverUptakeMeasurement/commit/fe48434a7df6898c741f31ec35b3536ec4eadbd4) listing the DCMQI dependency:
https://github.com/QIICR/PETLiverUptakeMeasurement/blob/fe48434a7df6898c741f31ec35b3536ec4eadbd4/CMakeLists.txt#L14

pieper:
Do you want to just convert this into a stub extension that just has a note that it's not needed anymore?  Since there is certainly documentation out there that says this is mandatory and it'll take a while for all such references to be updates.

DCMQI.json
jamesobutler, line 1: This file should actually be moved to the ARCHIVE directory of this repo per the README with a new entry added to the table explaining why it was removed.

https://github.com/Slicer/ExtensionsIndex/blob/main/ARCHIVE/README.md

fedorov:
Good point @pieper - are there any examples of such stub extensions? I don't think I've seen those before. I agree this would be better probably.

ARCHIVE/PETLiverUptakeMeasurement.json
fedorov: Good catch!

pieper:
I don't know of a case quite like this before, but I think you could probably just remove all the modules and it would just work.  Or, perhaps, you may need one module, so you could make one and set the hidden flag.  Then you could update the message on the extension manager that the module is no longer needed since the functionality is in core (I think that's just the README).

This would fix the issue James pointed out as well.

lassoan:
I don't think we need a stub - we can just remove the extension from the nightly (Slicer-5.11) branch. The main point in having the preview branch is to iron out any issues before the next stable release.

pieper:
I still vote for a stub.  Too much confusion otherwise.

pieper:
Maybe just keep the stub for one major release.

jamesobutler:
If there is a proposal to change ARCHIVE steps for extensions then updating the readme at https://github.com/Slicer/ExtensionsIndex/tree/main/ARCHIVE should be required. Otherwise the existing archive policy should be followed.

Previous stable releases such as Slicer 5.10 included mentions of archived extensions as well as new extensions in the community announcement of the new release. See https://discourse.slicer.org/t/slicer-5-10-summary-highlights-and-changelog/45224/2#heading--removed

pieper:
I agree @jamesobutler this seems to be a category of ARCHIVE that's not handled by the current guidelines.  I'm just pretty sure there are youtube videos and other tutorials out there that say this extension is required and people will be confused if they can't find it.  Maybe instead of just removing extensions from search results completely when they are archived, we could have the extension manager return a pointer to the `ARCHIVE.md` file section that explains why the extension was archived.  This could be a "tier -1" or something.

jamesobutler:
As it relates to older tutorials, doesn't the concept of a feature being removed like LightBox or module UI being reorganized for new features or logic updating to change a feature to act in a different way all things that can happen? Then in general for all of these situations where there is a change in the application, how else would we apply a policy that keeps it around or suggests "hey this button in this UI is no longer here and has been moved"? I think this extension case is the same as those other cases for Slicer core. At some point the tutorial viewer has to use the same version that the tutorial was made with to replicate step for step, or the tutorial creator needs to provide updates on an as-needed-basis. Our documentation for a given version isn't going to detail what is currently present and what also used to be present.

pieper:
Sure, we can't handle every situation but when it's not too hard to make it easier for users we can try.  Slicer rightly has a reputation as "hard to learn" so I'd suggest reasonable steps if we can.  If it's too much work don't bother and put our efforts into more effective things.

fedorov:
@lassoan I see you added the workflow that is failing just two weeks ago. Can you take a look at those failures? They do not make sense to me. I have not change the failing dependency.

lassoan:
Packaging fails with this error:

```
  Could not find a package configuration file provided by
  "SlicerDevelopmentToolbox" with any of the following names:

    SlicerDevelopmentToolboxConfig.cmake
    slicerdevelopmenttoolbox-config.cmake
```

The packaging test currently does not support extension dependencies, so this is not an error of the extension but the test itself. I'll merge this now.

lassoan (approved):
I've updated the packaging test, all tests pass now, all looks good.

PR #2354, opened 2026-05-15
Retire SlicerLiver from Slicer 5.8

RafaelPalomar:
Retires the `SlicerLiver` extension entry from this Slicer release branch.

The Slicer-Liver maintainers (ALive-research) have formalised a two-branch model in [ADR-0006](https://github.com/ALive-research/Slicer-Liver/blob/main/Docs/adr/0006-branch-model.md): `main` for Slicer-stable, `preview` for Slicer-preview / `main`.  The previously-named `master` branch was renamed to `main` in the same change set, aligning with 3D Slicer upstream's 2022 inclusive-language transition ([Slicer/Slicer#6277](https://github.com/Slicer/Slicer/pull/6277)).

The post-ADR-0006 `main` branch carries content (Qt 6 migration, modernised APIs picked up in `9ef780b COMP: Fix compile errors`) that does not reliably compile against this older Slicer release.  GitHub does install a permanent `master` → `main` ref redirect, so an entry with `scm_revision: master` would still resolve — but the resolved build fails.  Masking a known-incompatible state with a successful index entry would be worse than retiring the entry outright.

End-users on this Slicer release are encouraged to upgrade to Slicer stable (5.10+) to continue receiving Slicer-Liver.

---

If the Slicer team prefers to keep an entry pointing at a frozen legacy SHA on Slicer-Liver's side, the alternative is for Slicer-Liver to tag a pre-divergence SHA (e.g. `legacy-slicer-5.x-final`) and change `scm_revision` here to that tag.  Happy to do that variant if preferred.


lassoan:
Sorry, I just see this now. Removing an extension retroactively is not a good idea, as the extension packages are still present, so it would just become harder to find the source code. The cleanest solution if SlicerLiver developers create a 5.8 branch (from the latest commit that is still compatible with Slicer-5.8) and the extension description is updated to use that branch.

PR #2357, opened 2026-05-25
Add Radpretation extension

bhavyradpretation:
# New extension

## Overview

Radpretation Plugin is a 3D Slicer extension focused on Orthanc PACS integration, DICOM workflow automation, and remote radiology workspace interaction.

The extension enables seamless communication between Orthanc PACS and 3D Slicer, allowing users to load studies, perform segmentation workflows, and synchronize imaging data efficiently within clinical and research environments.

## Main Features

- Orthanc PACS integration
- Automated DICOM study loading via remote trigger
- DICOMweb support (QIDO-RS / WADO-URI)
- Seamless push/pull of DICOM data between Slicer and PACS
- One-click DICOM Segmentation (SEG) upload workflows
- Remote imaging workspace interaction (Local Bridge Server)

## Notes

- The extension uses secure user-controlled PACS communication.
- No patient data is transmitted externally without user action.
- Designed for Orthanc PACS and OHIF-based workflows.
- Segmentation synchronization enhancements are planned in future releases.

## Tier 1 Checklist

- [x] Extension has a reasonable name
- [x] Repository name follows convention
- [x] Repository has `3d-slicer-extension` topic
- [x] Extension description added
- [x] LICENSE.txt added
- [x] Extension URL and revision are correct
- [x] Screenshot URLs are correct
- [x] Homepage URL is valid
- [x] Does not send information without user consent

## Additional Comments

This extension is currently under active development. Future updates will include enhanced segmentation synchronization, workflow automation, and improved DICOM interoperability support.


pieper:
This looks nice 👍 

It would be nice to have a brief tutorial for this when you have time.  But that's not a blocker since this sounds pretty much ready to go.

bhavyradpretation:
Thank you for the review and feedback!
Glad to hear the extension looks good.

I’ll add a brief tutorial/documentation for the workflow and setup in an upcoming update to make onboarding easier for users.

Appreciate your time reviewing the extension.

lassoan:
Thank you for your submission. It is an interesting extension, which addresses some of the pain points of 3D Slicer.

One thing that would be important to improve if you want the extension to be discoverable and usable is to describe more clearly what it is for. Very hard to figure out what this extension does, even after reviewing the code, I have just a faint idea. I guess it implements the common workflow of pull images, segment them, and push back to a server, but it is not clear how it can be used. Do you need a custom server (Radpretation server)? What can you do with that server? What can you do without that server? Screenshots could help, but there is just one (ss.png) and it is not very informative; and it does not appear in the README.md file, so most people would not find it anyway.

I'm just curious: Why Orthanc-specific REST API is used instead of DICOM networking? (you don't want other DICOMweb servers to be compatible with the extension or there are significant advantages of the REST API?)

Suggestion to consider for CacheManager: it may be better to get temporary folder or cache folder from slicer.app instead of using system default folder, as user may prefer to use a certain drive for temporary files (and manual cleanup of leftover files is easier if they are in a known location).

Most of the features are already available in Slicer core and extensions, but your implementation has some advantages compared to them. We ideally would like community members to improve existing modules instead of coming up with parallel implementation, for best use of everyone's time. Were you aware that DICOM and WebServer modules in Slicer core and DICOMWeb extension provide similar features? Have you considered improving them instead of coming up with a new implementation? Would you be still open to consider improving these existing modules as much as possible to minimize duplication? This is not a requirement for accepting your extension, just trying to see if it was possible to join forces instead of working in parallel.

PR #2394, opened 2026-09-15
Add AGENTS.md to file allow list

ebrahimebrahim:
Adding AGENTS.md to file allow list, so that the validate-extensions workflow can pass


pieper (approved):
Thanks 👍

PR #2397, opened 2026-09-17
Add OMEZarr extension

vboussot:
# New extension

OMEZarr opens and saves OME-Zarr (OME-NGFF) images in Slicer: drop a local
`.ome.zarr` folder or paste an `https://` / `s3://` address. Large images open
at a resolution that fits in memory, and the slice views can then be refined
to full resolution, reading only the chunks they need. It is built on
[ngff-zarr](https://github.com/fideus-labs/ngff-zarr), which it installs from
PyPI on first use.

Repository: https://github.com/fideus-labs/SlicerOMEZarr

The module self-test runs in GitHub Actions against the latest stable Slicer
on Linux, macOS and Windows.

## Tier 1

- [x] Extension has a reasonable name
- [x] Repository is associated with `3d-slicer-extension` GitHub topic
- [x] Extension description summarizes in 1-2 sentences what the extension is usable for
- [x] Any known related patents must be mentioned in the extension description (none)
- [x] LICENSE is present in the repository root (MIT) and the license is named on the homepage
- [x] Extension URL and revision are correct (`main` branch)
- [x] Screenshot URLs are correct, contains at least one
- [x] Content of submitted json file is consistent with the top-level CMakeLists.txt
  - [x] Extension name
  - [x] Short description
  - [x] At least one nice, informative image (a walkthrough video and screenshots)
  - [x] Description of contained modules
  - [x] Publication: none yet
- Hide unused github features:
  - [x] Wiki, Projects and Discussions are disabled
  - [x] Releases and Packages are unchecked in About
- The extension is safe:
  - [x] Does not send any information anywhere without user consent


lassoan:
We have an [advanced-stage prototype branch](https://github.com/lassoan/Slicer/tree/voxel-value-scaling) that adds support for OMEZarr to Slicer natively, achieving a similar feature set that this extension, by making several Slicer core use a voxel data provider class instead of working from a in-memory volume. Please have a look and let's brainstorm on how we could make this feature set available in Slicer the best possible way.

vboussot:
@lassoan Thanks, I hadn't seen this branch. I had only looked at branches and pull requests in the main Slicer repository, not forks.

I also had a small list of changes I wanted to propose for Slicer core (https://github.com/fideus-labs/SlicerOMEZarr/issues/1), and your branch already covers some of them. Really nice work on this.

I went through the branch. The display side is something an extension can only approximate: real multiresolution in slice views and volume rendering, with caching and a memory budget.

What is still missing is mostly the format side, which is what ngff-zarr maintains: Zarr v3 and OME-Zarr 0.5+, axis units (a store in micrometres currently comes in 1000× too large), RFC-4 orientation, channels, time series, labels, S3, and writing.

The extension complements this by providing that format support on released Slicer versions today, but it is Python, volumes still have to fit in memory, and the full-resolution overlay only approximates what your branch does natively.

I think the best path is to use the extension in the meantime, while progressively moving these capabilities into Slicer core. I'm happy to test the branch and remove parts of the extension as they become available natively.

Longer term, if the format side can also be integrated cleanly, it would be great to have the whole stack natively in Slicer.

Would it make sense to list the extension as experimental for now?

Also looping in @thewtex.

lassoan (approved):
I've reviewed the extension and it looks ready to go. When voxel provider feature will be added to Slicer core then some of the module features no longer be needed, but there is no reason to hold up distribution of this extension.

thewtex:
I fully agree with Valentin's comments / assessment. The native loading for rendering is a good eventual target, and Andras does have excellent initial progress there. However, the standard is much more than loading the pixel buffer. And even loading the pixel buffer correctly depends on the rest of the specification. There any many important pieces encountered in real-world usage. And there are many critical 3D Slicer integration pieces like spatial metadata, orientation. ngff-zarr has ~2,000 tests covering the importing pieces and is evolving with the specification.

> I think the best path is to use the extension in the meantime, while progressively moving these capabilities into Slicer core.

:+1: 

> Longer term, if the format side can also be integrated cleanly, it would be great to have the whole stack natively in Slicer.

:+1:

thewtex:
> I've reviewed the extension and it looks ready to go. When voxel provider feature will be added to Slicer core then some of the module features no longer be needed, but there is no reason to hold up distribution of this extension.

@lassoan :100: :pray:

lassoan:
Is there a C++ implementation that may be usable or you would recommend Slicer core to call out Python libraries for NGFF I/O? C++ would have the advantage of less complications with multithreading and webassembly builds.

thewtex:
I think it may work for now to call out to Python for NGFF I/O until there is a C++ interface library that servers Slicer's needs. And work towards the transition Valentin suggested.

There will not be issues with webassembly.

For multithreading, the Python GIL will need to be considered. In the future free-threaded Python may help, but that is not a practical option currently.

